PR #109, opened 2014-10-03
Pims bioformats

caspervdw:
This is a wrapper for python-bioformats. Required are:
- https://github.com/CellProfiler/python-bioformats
- manually overwrite loci_tools.jar with newest version
- https://github.com/CellProfiler/python-javabridge
   or (windows compiled) http://www.lfd.uci.edu/~gohlke/pythonlibs/#javabridge   
- Java JDK

Unless you are not planning to work with big files, 64bits everything is required.

Big files still have long loading times because metadata of all frames is read in the beginning. Maybe it is better to put all metadata routines separate from get_frame and _initialize, then the user can choose. A lot more metadata can be pulled from _jmd, see http://downloads.openmicroscopy.org/bio-formats/5.0.5/api/loci/formats/meta/MetadataRetrieve.html

get_frame results in a 3D image (if _dimension = 3), its index is the time index, its channels are set by the property channel and the multipoint index by the property multipoint. You can force it to 2D, then the index is the plane index. 

There is a strange issue with nd2 files: sometimes CTZYX images appear as multipoint CTYX images. The code checks for a suspicious file and then automatically goes to a mode that corrects for this issue ('use MP as Z')


caspervdw:
Hi guys,
It has been silent in this PR! To finish this, I made some unittests for pims-bioformats, on file types that are not too large to put in pims. To gain some testing time, I temporarily removed pyAv. When I include pyAv again it says:

    /home/travis/build.sh: line 41: 20138 Segmentation fault      nosetests

I don't get this, some conflict between pyav and jdk/javabridge/bioformats? Or maybe Travis is limited in memory space in some way.. Do you have any idea what may be going on here?

Finally, I still have some ideas to improve on the metadata reading routines. I now switched this off because java was printing a lot of exceptions that clobbered Travis log, I will switch it back on when I have figured out how to suppress these (https://github.com/CellProfiler/python-javabridge/issues/37).

tacaswell:
`export LD_LIBRARY_PATH=/usr/local/lib LIBRARY=ffmpeg` this is almost certainly the issue.  The issue is almost certianly a library version cross between the system libraries and the conda libraries.

caspervdw:
If it's a library cross between system and conda, then why would the master work? I think this is a library conflict between javabridge and ffmpeg. I tried some alternatives to installing ffmpeg (via `sudo apt-get`) but it's unsupported on Ubuntu 12.04. That's why Dan used this script, probably.

@danielballan, any ideas?

danielballan:
Yes, I used the script because 12.04 doesn't offer the right version of ffmpeg (or, to be precise, its underlying libraries) for PyAV. The script is more or less lifted from the build script that PyAV itself uses to run tests on Travis. The problem we are having is probably due to conflicting versions of libc.

I have sunk many hours of effort into making cross-platform conda packages for ffmpeg and PyAV in effort to clean this up. I have some concrete progress, but no package yet, and it's not clear yet whether that can be a priority for me in my new job.

In the short term, I think our users will have to use javabridge *or* PyAV, never both concurrently. For testing purposes, let's configure Travis to run a separate build that does not include PyAV but does run javabridge.

Casper, are you able to do this? You can use trackpy's testing matrix (.travis.yml) as a guide. If you are too busy or not interested, I can pick this up next week.

caspervdw:
Thanks for the response, I fixed Travis in the way you suggested. It appears that since https://github.com/soft-matter/pims/pull/126, pyav unittests were skipped. Sorry about that. This had to do with the `rm libm.so.6` that I deleted. 

I finally got a clear picture of all the metadata reading options, it's now in far better shape than it was. The options are described in the docstrings.

The issue is that at initialization of MetadataRetrieve (so at initialization of the reader), java prints a lot of `Exception in "Thread-0"` messages to the console. You don't see them in IPython. I can't stop them as they bypass the java `system.err` and `system.out`, which I redirect to `frames.java_log`.

caspervdw:
The bioformats extension has been rebased and it is roughly finished. It is only waiting on merge of PR #131, if it is merged then the bioformats tests will probably pass.

Apart from the bioformats reader, I've added a priority handling in `pims.open`, so that bioformats doesn't get executed before more conventional tiff readers. It is quite basic, it looks for a `class_priority` attribute which defaults to 10. I set the priority of Bioformats to 5, so that other readers always get chosen before Bioformats.

I hope this can be merged in the near future.

caspervdw:
As discussed in #131, I removed all binaries but the nd2 file of 500 kB. This file is used for testing the metadata reading.
Instead, there now is a download script that allows the user to download and test with several filetypes.

danielballan:
nosetests is picking up `download_bioformats_test.py`, and Travis is failing. Edit `.travis.yml` so that it runs `nosetests pims/` instead of `nosetests .`, and we should be golden.

As soon as Travis passes, we'll merge. This doesn't touch any existing code except the (previously implemented) reading priority, and I gather you've been using it successfully for your work for some time. It's time to get this into master.

caspervdw:
It should pass now, I renamed the function instead.

danielballan:
Travis is clogged up today....

pims/bioformats.py
```diff
@@ -0,0 +1,692 @@
...
+    try:
+        import javabridge
+        import bioformats
+    except ImportError:
+        return False
+    else:
+        return True
+
+
+class MetadataRetrieve():
```
tacaswell: this should inherit from `object`

caspervdw: will change

caspervdw:
@danielballan, you mean the `Exception in thread 0` prints? I can't figure out a way to suppress these. I've opened an issue on python-javabridge (https://github.com/CellProfiler/python-javabridge/issues/37). I already redirected the java `stdErr` and `stdOut`.

danielballan:
@caspervdw No, haha, I was just griping that Travis was taking forever to run your tests.

PR #114, opened 2014-10-27
ENH : added sprite sheet class

tacaswell:
Wraps a sprite sheet (MxN grid of rxc images) with a pims
class.

Test are forthcoming, but it is dinner time.


tacaswell:
This went into the docs in https://github.com/soft-matter/pims/blob/master/doc/source/custom_readers.rst#example-demonstrating-generality-of-pims-design

PR #118, opened 2014-11-25
ENH: Multidimensional image-sequence

caspervdw:
`ImageSequence3D` class derived from `ImageSequence`.
Imports t, z, c indexed .tif files. Output is indexed with t index and can be 2/3/4 dimensional.


danielballan:
Great, thanks. Sorry for the long delay on this. Please open a new issue reminding us to add tests on `ImageSequence3D`, which we can address once we have a public solution for storing more binary data.

PR #127, opened 2014-12-23
Natural sorting of file names in ImageSequence

### LINKED ISSUE #103: natural sorting of file names in ImageSequence
Implement something like this is image sequence because there exists software without left-padding.

http://stackoverflow.com/questions/4836710/does-python-have-a-built-in-function-for-string-natural-sort.

hadim:
Should fix #103.

### DIFF
--- a/pims/utils/sort.py
+++ b/pims/utils/sort.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+import re
+
+__all__ = ["natural_keys"]
+
+
+def _atoi(text):
+    return int(text) if text.isdigit() else text
+
+
+def natural_keys(text):
+    """Sort list of string in a human way.
+    See: http://stackoverflow.com/questions/5967500/how-to-correctly-sort-a-string-with-a-number-
+    inside
+    Examples
+    --------
+    >>> alist=["something1", "something12", "something17", "something2"]
+    >>> alist.sort(key=natural_keys)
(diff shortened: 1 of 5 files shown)


caspervdw:
Only thing from my side: I would put the new function in `image_sequence.py` directly. I think this will only be used for image sequences.

pims/utils/sort.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+from __future__ import division
+from __future__ import absolute_import
+from __future__ import print_function
```
caspervdw: I prefer the four on one line. Or are there reasons to split them?

tacaswell: No reason to spilt them (but they don't all fit on one line and comply with
pep8)

danielballan: Thanks, @hadim.

I would prefer

    from __future__ import (absolute_import, division, print_function,
                            unicode_literals)

if you have time to push a quick revision.

This might be useful for custom readers, so I am OK with putting it `utils` where others can import it for use in their own readers.

hadim: Sorry to be long. I'll try to push this next week !

hadim:
Clean rebase. Should be ok to merge.

danielballan:
Travis failed on the py2 tests. Looks like you need to add `'pims.utils'` to the list of packages in `setup.py`.

hadim:
Travis is happy now.

danielballan:
Great! Thanks for your contribution.

PR #129, opened 2015-01-17
FIX Travis pyav installation

caspervdw:
Fixes failed travis installation after PR #126.


.travis.yml
```diff
@@ -1,28 +1,40 @@
...
+        wget http://repo.continuum.io/miniconda/Miniconda-3.7.3-Linux-x86_64.sh -O miniconda.sh; 
+    else 
+        wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh; 
+    fi
   - chmod +x miniconda.sh
   - ./miniconda.sh -b
   - export PATH=/home/travis/miniconda/bin:$PATH
   - export PATH=/home/travis/miniconda3/bin:$PATH
   - conda update --yes --no-deps conda
+  - conda update --yes conda
```
tacaswell: Isn't this or the line above redundant?

caspervdw: This had something to do with an issue described here:
https://groups.google.com/a/continuum.io/forum/#!topic/conda/RSFENqovnro

It's the same code as in the trackpy travis. But it appears that it isn't necessary (anymore) so I will just delete the `conda update --no-deps`.

caspervdw:
It appears that `conda update conda --no-deps` wasn't necessary anymore; Travis passes when I delete that line. 

Just as in trackpy, I changed the python 3 version to 3.4.

I added some fixes for testing and rearranged `test-common.py` somewhat, to easily merge in the tests for bioformats and imagesequence3d. No major changes.

.travis.yml
danielballan: I think it would be better to leave that in. Currently, I agree it does nothing for us, but that issue could arise again as the hard-coded version of conda becomes stale.

caspervdw: I agree, I've put it back in

tacaswell: I don't understand why we do a no deps update before doing the same update
with deps. I assume there is some issue I am not aware of?

danielballan: Official conda sometimes updates its deps such that they are incompatible with older versions of conda. This protects us from that scenaio. See the issue Casper linked, in which I hashed this out with Aaron and Ilan.

tacaswell: Ah, sorry, was reading this in email and missed that link.

caspervdw:
Can we merge this one? Then I can rebase Bioformats and the ImageSequence3D tests.

PR #131, opened 2015-01-25
Add tests for ImageSequence3D

caspervdw:
Here are some tests for ImageSequence3D. I separated greyscale tests from the rest so that it can only be included for 2D RGB images. For 3D it raises a `NonImplementedError` and for non-RGB it does nothing, I don't think it is necessary to test that.

Also I added ImageSequence3D to pims api and made some docstrings for it.


danielballan:
I generally want to postpone putting binary data in, but these files are small and testing this is important, so let's do this. @tacaswell if we agree with that judgement call please merge.

tacaswell:
*grumble grumble*

It looks like this only adds ~4M, we already have a bunch of images, and this does seem like the minimal number of frames needed.

That said, the interesting thing we are testing here the logic in the names/data reshaping.  I think we can trust pillow to _write_ out valid pngs and do all of these tests with synthetic data.  Use the setup/tear down methods to create and delete the synthetic data.

I don't see any show-stopper problems going that route and would like to be convinced there is before we merge in more binary files.

caspervdw:
The pngs plus the numpy files are only about 250kB altogether. Is this amount really a problem?

If it is, I think the same goes for the TIFF tests. `stuck.tif` + npy files take around 3.5 MB. So if we are looking for size reduction, we could also make a synthetic TIFF file and save a lot more.

The filetypes that cannot be generated, we could crop to say 10x11 pixels. The Bioformats tests #109 have around 7 MB of binaries. I haven't figured out how to crop these files to get them smaller but I am sure it is possible.

danielballan:
@tacaswell I still vote to merge, unless you are volunteering to write the code to generate this data programmatically. Forgive the cheekiness. :- )

@caspervdw Removing `stuck.tif` would decrease the size of a pip/conda install, but it would not decrease the size of the repo (which of course contains the whole history). Unfortunately, I committed those files in the early days before I knew what I was doing. The only way to remove them would be to force-push to master, which is a mortal sin. So, as long as they are in there, we use them.

danielballan:
One more thought: does scikit-image have good tools for generating images? I know they also have some binary images in their examples.

caspervdw:
I see, well I think only reducing the size of the pip install is also nice, good to keep PIMS as light as possible. I believe the repo limit is around 1GB?

As far as I know, scikit-image uses PIL or tifffile to write images.

Concerning bioformats, I could try retrieving them from http://loci.wisc.edu/software/sample-data during test, but this could be unstable. There are a lot of formats in there, some of them several 100MB, and some of them giving errors.

caspervdw:
So I'll write some artificial data generating `setUp`, ok? Shall I implement it in the normal `ImageSequence` too?

danielballan:
That would be great.

I don't think we want to wait for a ~100 MB data download in Travis at this time. Let's write the tests in a way that Travis will skip them but we can run them locally. Here is my suggestion: create a Python script at the top level like `download-test-data.py`. In the tesitng suite, configure every test that requires downloaded files to raise `SkipTest` if the file it requires is missing.

caspervdw:
I suggest including only 1 Bioformats file for travis testing, I can produce a really small nd2. Let's continue this discussion in #109.

caspervdw:
I just force-pushed the artificial png-generating script, now the previously added binaries are out of the repository.

danielballan:
Very good. Thanks for cleaning up my old mess. So, to summarize, there are now no binary packages involved in testing ImageSequence and ImageSequence3D.

This looks ready to merge. On to #109....

PR #133, opened 2015-01-27
FIX Display function interprets `colors` metadata

caspervdw:
It was already implented but still buggy. The rich display function now correctly interpret `colors` field in metadata dictionary.

Additional fix: PIL doesn't take RGB images with RGB axis at a position other than the inner position (tested with (3, 256, 512) shaped array). So I changed the definition of `has_color_channels`. This has the advantage that depth-3 and depth-4 z-stacks get viewed correctly.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -311,7 +311,7 @@ def to_rgb(image, colors=None, normalize=True):
         datatype is np.uint8, else it is float.
     """
     # identify whether the image has a (leading) channel axis
-    if colors == None:
+    if colors is None:
         has_channel_axis = image.ndim > 2 and image.shape[0] < 5
     else:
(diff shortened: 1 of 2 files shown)


caspervdw:
I have been using this code for a while now and it seems a good idea to merge.

pims/display.py
```diff
@@ -310,9 +310,13 @@ def to_rgb(image, colors=None, normalize=True):
         so that values lay between 0 and 255. When normalize = True (default),
         datatype is np.uint8, else it is float.
     """
+    # identify whether the image has a (leading) channel axis
+    if colors == None:
```
tacaswell: `if colors is None` please

caspervdw: Thanks! I corrected it.

PR #137, opened 2015-02-06
ENH: Bioformats crop + minor fix

caspervdw:
This enhancement makes use of the built-in cropping methods of bioformats. Useful for very big images. It works like this:

    frames = pims.frames(filename)
    frames.crop = (20, 20, 100, 100)  # get a 100x100 frame with origin 20, 20
    assert frames[0].shape == (100, 100)

Additionally, there is a minor fix for metadata reading with bioformats 5.1.x.

### DIFF
--- a/pims/bioformats.py
+++ b/pims/bioformats.py
@@ -70,6 +70,13 @@ class MetadataRetrieve(object):
                 return None
             # convert value to int, float, or string
             jw = str(jw)
+            try:  # deal with values hidden inside 'value[number]'
+                temp = jw[jw.index('value[') + 6:]
+                temp = temp[:temp.index(']')]
+            except ValueError:
+                pass  # do nothing when 'value[' or ']' were not found
+            else:
+                jw = temp
             try:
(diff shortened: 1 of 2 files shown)


pims/bioformats.py
```diff
@@ -343,6 +353,13 @@ def _change_series(self):
         else:
             self._first_frame_shape = (self._sizeY, self._sizeX)
             self._sizeC = self.rdr.getSizeC()
+            
+        # Reassert crop area
+        crop_area = self.crop
+        try:
+            self.crop = crop_area
+        except ValueError:
+            self._crop = None
```
tacaswell: Should probably use the property here too to be consistent.

tacaswell:
This is another feature I think should be pushed up the stack and implemented on all sequences (done via brute-force by default).

caspervdw:
I agree, additionally I am busy on another (much faster!) implementation of Bioformats reader. I'll close this and look at the crop again later.

PR #138, opened 2015-02-15
ENH: Image sequence in zipfiles

caspervdw:
When a path to a zipfile is specified, all files inside the zipfile will be loaded as an imagesequence using `PIL.Image.open`. For large collections of images, it may be preferred to work from zipfiles.

### DIFF
--- a/pims/image_sequence.py
+++ b/pims/image_sequence.py
@@ -5,8 +5,11 @@ import six
 from six.moves import map
 import os
 import glob
+import fnmatch
 from warnings import warn
 import re
+import zipfile
+from StringIO import StringIO
 
 import numpy as np
 
@@ -16,6 +19,7 @@ from pims.utils.sort import natural_keys
 
 # skimage.io.plugin_order() gives a nice hierarchy of implementations of imread.
 # If skimage is not available, go down our own hard-coded hierarchy.
+from PIL import Image
 try:
     from skimage.io import imread
 except ImportError:
@@ -32,9 +36,10 @@ class ImageSequence(FramesSequence):
     Parameters
     ----------
     path_spec : string or iterable of strings
-       a directory or, safer, a pattern like path/to/images/*.png
-       which will ignore extraneous files or a list of files to open
-       in the order they should be loaded.
+        a directory or, safer, a pattern like path/to/images/*.png
+        which will ignore extraneous files or a list of files to open
+        in the order they should be loaded. When a path to a zipfile is
+        specified, all files in the zipfile will be loaded.
     process_func : function, optional
         callable with signalture `proc_img = process_func(img)`,
         which will be applied to the data from each frame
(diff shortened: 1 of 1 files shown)


pims/image_sequence.py
```diff
@@ -16,6 +19,7 @@
 
 # skimage.io.plugin_order() gives a nice hierarchy of implementations of imread.
 # If skimage is not available, go down our own hard-coded hierarchy.
+from PIL import Image
```
tacaswell: I always get confused with the PIL/pillow conventions.  Does this work with both?  We should prefer pillow to PIL.

pims/image_sequence.py
```diff
@@ -102,6 +121,17 @@ def _get_files(self, path_spec):
             self._count = len(path_spec)
             return
 
+        if zipfile.is_zipfile(path_spec):
+            self.is_zipfile = True
```
tacaswell: For readability reasons can you put a `self._is_zipfile = False` and `self._zipfile = None` in the `__init__`?  I also suggest making these 'private' attributes as these details should be mostly hidden from the users.

tacaswell: Can you put either an exception or a warning in here if the user passes both a zipfile _and_ a plugin?  Silently ignoring user input is a bad thing.

tacaswell:
Awesome!

I left a few style comments, but other wise looks good to me.

pims/image_sequence.py
caspervdw: I tested it with Pillow only, but I think `Image.open` has the same syntax in both versions.

tacaswell: Great, then disregard this comment.

caspervdw:
I implemented the changes you suggested and additionally made it compatible with Python 3.x . I think it can be merged.

pims/image_sequence.py
```diff
@@ -94,6 +101,20 @@ def __init__(self, path_spec, process_func=None, dtype=None,
         else:
             self._dtype = dtype
 
+    def close(self):
+        if self._is_zipfile:
+            self._zipfile.close()
+
+    def __del__(self):
```
tacaswell: Can you push this up to the base-class (along with the context manager stuff)?

pims/image_sequence.py
```diff
@@ -94,6 +101,20 @@ def __init__(self, path_spec, process_func=None, dtype=None,
         else:
             self._dtype = dtype
 
+    def close(self):
```
tacaswell: Probably worth using a `super` call in here to get the base-class close.

tacaswell:
Sorry, two more minor things, but looks great other wise!

pims/image_sequence.py
caspervdw: I am not sure if it's a good idea to define the destructor to call `close` in all cases. If you call `frames.close()`, and then destroy the object, `close()` will be run twice which might give trouble for some readers.

For the zipfile it's no problem to close it twice, that's why I put it in for this specific reader.

`__enter__` `__exit__` already works by defining `close()`, right?

tacaswell:
Thanks!

PR #140, opened 2015-03-10
Matplotlib plot to Frame conversion

caspervdw:
This adds `normalize`, `to_rgb`, `plot_to_frame`, and `plots_to_frame` to the root `pims` namespace.

- `normalize` is renamed from `display._normalize`, for this all parameters that were named `normalize` are now named `normed` (I believe that in numpy, they also use `normed` for histograms, not `normalize`)
- `to_rgb` was already there in `pims.display.to_rgb`
- `plot_to_frame` renders a mpl Figure or Axes to an RGBA image, wrapped by `pims.Frame`
- `plots_to_frame` does the same, but for an iterable of Figures / Axes objects. It outputs a 3D Frame, to display as a scrollable stack. Possibly this route may be used for easy video output of matplotlib figures via PyAv (but I don't have PyAv so I can't test that)

This PR is needed for `trackpy.plots.annotate3d` (https://github.com/soft-matter/trackpy/pull/196)


pims/matplotlib.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import (absolute_import, division, print_function,
```
tacaswell: shadowing `matplotlib` is probably technically ok, but I think is bad style.

I would also import mpl at the top of this file and not import these as part of the api (or put this import in a try-except)

caspervdw:
A response on @tacaswell line note: I was looking for a method to transform any iterable of axes objects (produced by any tp.plot function) into something that is displayed nicely in IPython. This is the best I could come up with. Additionally, this could be used for exporting movies.

I would like to solve this as 'matplotlib' as possible. It's just the best I could do. Do you have any idea of other approaches?

danielballan:
I think @tacaswell was concerned about the name of the file colliding with `matplotlib` itself. We could call it `mpl_conversion`, `mpl_utils`, or whatever you like.

caspervdw:
Ah sorry I misunderstood his comment. As it is only intended for display
applications, we could put it in `display.py`. What do you think?

tacaswell:
I think that is a better name, but I have a bad track record at
intentionally naming things.

danielballan:
Haha. Go for it Casper. Stay away from the names, Tom. :- D

caspervdw:
Moved.
Does anyone understand what is going wrong in PIL? It's saying `ImportError: No module named PIL`. I tried some things:

- `conda create image` instead of `pillow`, gives package not found
- `pip install pillow` eventually gives `ImportError: No module named scipy.ndimage` (https://travis-ci.org/caspervdw/pims/builds/55486229)
- `pip install scipy` scipy installation now fails, it misses numpy (https://travis-ci.org/caspervdw/pims/builds/55486317)

It seems like packages are not properly installed. Any thoughts?

caspervdw:
It's now fixed and ready for merge.

pims/display.py
```diff
@@ -349,6 +365,78 @@ def to_rgb(image, colors=None, normalize=True):
     result = result.clip(0, 255)
 
     if normalize:
```
tacaswell: Missed a renaming.

pims/display.py
```diff
@@ -158,7 +163,7 @@ def repr_video(fname, mimetype):
     return HTML(data=video_tag)
 
 
-def _scrollable_stack(sequence, width, normalize=True):
+def _scrollable_stack(sequence, width, normed=True):
```
tacaswell: Are we not worried about the API change here?

caspervdw: As it is an unreleased and private (_) function I think this should be ok.

danielballan: Fine by me

caspervdw:
I did an additional fix that effects the same code. Ready for merge.

PR #144, opened 2015-04-09
Bioformats jpype

caspervdw:
I had some persistent issues with the Bioformats reader having to do with very slow reading times and crashes on large files. So I tried out another java interface `jpype` and it appears to be approx 20x faster and easier to use. Additionally, it will probably be going to support Py3 soon. And it doesn't conflict with PyAv.

Some minor changes:
- MetadataRetrieve are now called without `get`, i.e. `frames.metadata.PlaneTheC(0,0)` instead of
`frames.metadata.getPlaneTheC(0,0)`
- For some reason, pictures do not get recognized as RGB but (for all my tests) as 3-channel picture. I think this is an improvement.
- kill_vm or, with JPype, ShutdownJVM is not necessary as this is called automatically on python exit
- loci_tools.jar is automatically downloaded when reader is called and loci_tools.jar is not found in pims root folder

### DIFF
--- a/pims/bioformats.py
+++ b/pims/bioformats.py
@@ -12,32 +12,69 @@ try:
 except ImportError:
     jpype = None
 
-LOCI_TOOLS_PATH = os.path.join(os.path.dirname(__file__),
-                               'loci_tools.jar')
 
 def available():
-    return jpype is not None 
+    return jpype is not None
 
 
-def download_jar(url=None, overwrite=False):
-    """Downloads loci_tools.jar from openmicroscopy into the pims root folder.
-
-    Parameters
-    ----------
-    url: string
-        specifies a custom loci_tools.jar url, for instance to change version
-    overwrite: boolean
-        set overwrite=True to overwrite the existing loci_tools without notice
+def _gen_jar_locations():
     """
+    Generator that yields optional locations of loci_tools.jar.
+    The precedence order is (highest priority first):
+
+    1. pims package location
+    2. /etc/loci_tools.jar
+    3. CONDA_ENV/etc/loci_tools.jar
+    4. PROGRAMDATA/pims/loci_tools.jar
+    5. LOCALAPPDATA/pims/loci_tools.jar
+    6. APPDATA/pims/loci_tools.jar
+    7. ~/.config/pims/loci_tools.jar
+    """
+    yield os.path.dirname(__file__)
+    yield '/etc'
+    if 'CONDA_ETC_' in os.environ:
+        yield os.environ['CONDA_ETC_']
+    if 'PROGRAMDATA' in os.environ:
+        yield os.path.join(os.environ['PROGRAMDATA'], 'pims')
+    if 'LOCALAPPDATA' in os.environ:
+        yield os.path.join(os.environ['LOCALAPPDATA'], 'pims')
+    if 'APPDATA' in os.environ:
+        yield os.path.join(os.environ['APPDATA'], 'pims')
+    yield os.path.join(os.path.expanduser('~'), '.config', 'pims')
+
+
+def _find_jar(url=None):
+    """
+    Finds the location of loci_tools.jar, if necessary download it to a
+    writeable location.
+    """
+    for loc in _gen_jar_locations():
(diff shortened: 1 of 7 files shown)


pims/bioformats.py
```diff
@@ -5,32 +5,53 @@
...
+    import jpype
 except ImportError:
-    javabridge = None
+    jpype = None
 
-try:
-    import bioformats
-except ImportError:
-    bioformats = None
+LOCI_TOOLS_PATH = os.path.join(os.path.dirname(__file__),
```
tacaswell: This should probaly use one of the XDG paths.  This works for conda or develop installs as you have write permission to the install location, but if pims is installed system wide this directory could be read-only.  This should probably also be a search cascade which leads to a rather annoying type of pain, see https://github.com/NSLS-II/metadatastore/blob/master/metadatastore/conf.py

caspervdw: I don't really like these search cascades and for simplicity would prefer putting it inside the pims folder. This way it seems to get pretty complicated.

What do you think of downloading it as part of `setup.py` (if jpype is there), or else suggest the user to download it to pims folder by himself?

tacaswell: I think that downloading at run time is the right approach (to prevent any strange installation order dependencies (install pims before jtype, you have both, but can't use bioformats)).

It is user hostile even by my standards to ask the users to download and copy files to a location we can't tell them directly.

It may be complicated, but it is the cost of doing business here.

caspervdw: Ok I'll work on some search cascade. It's not that I don't like it because it's complicated, I just don't like it because complicates often means more prone to errors.
But I'll start form the metadatastore conf file you linked.

caspervdw:
I fixed the save location of loci_tools.jar. Additionally, jpype 0.6.0 just came out, which supports python 3. I made a few changes so that this reader is now fully Py3 compatible :)

danielballan:
Woo-hooo! Good to merge?

pims/bioformats.py
```diff
@@ -1,36 +1,102 @@
...
+    2. /etc/loci_tools.jar
+    3. CONDA_ENV/etc/loci_tools.jar
+    4. PROGRAMDATA/pims/loci_tools.jar
+    5. LOCALAPPDATA/pims/loci_tools.jar
+    6. APPDATA/pims/loci_tools.jar
+    7. ~/.config/pims/loci_tools.jar
+    """
+    yield os.path.dirname(__file__)
+    yield '/etc'
+    if 'CONDA_ETC_' in os.environ:
```
tacaswell: This isn't a canonical thing, this is needs https://github.com/NSLS-II/conda-prescriptions/tree/master/releases/conda_etc to work.

pims/bioformats.py
```diff
@@ -1,36 +1,102 @@
...
+    return jpype is not None
 
 
-def available():
-    try:
-        import javabridge
-        import bioformats
-    except ImportError:
-        return False
+def _gen_jar_locations():
```
tacaswell: :+1: This is much better than they way I did it!

caspervdw: Haha I kind was really glad with the idea of a generator too ;)

pims/bioformats.py
caspervdw: So remove it?

caspervdw:
Before merging I just want to wait until the JByte[] -> np.ndarray conversion is clears up a bit (see previously referenced issue on JPype repo)

pims/bioformats.py
tacaswell: I think so, we are using it, but it will hopefully be superseded by something 'official'

tacaswell: 'we' here being day job.

caspervdw:
All should be OK and ready for merge now. On my system jbytearr[:] does not give a numpy array, but on Travis it does work apparently. Seems to have something do with my platform. I included some fallback functions (readout via string buffer and readout via java casting) to make it work on windows sytems. They will probably be deprecated soon when the issue has been fixed, but it will enable use of older JPype versions. It prints a warning message when one of these fallback functions are used.

danielballan:
Great. Here goes!

Issue #141, opened 2015-03-10
pims.TiffStack doesn't fetch the right content/values


forevertheuni:
Hello all, Thanks for pims

So. I am having a problem with pims. When I open a TiffStack the values of the pixels of an image do not correspond to the correct values:
Here's a bit of code to test it (PIL and libtiff work ok, meaning the same values as acquired by the microscope software and ImageJ):
Sample file can be dled here:
https://drive.google.com/file/d/0B9-erPeWaEjGdy10bFMteHdNXzQ/view?usp=sharing

Help would be much appreciated since I already wrote a bunch of functions relying on pims. 
```python
import numpy as np
import libtiff
import pims
from PIL import Image
file = 'TRIMcapture_MG132_06_R3D_D3D_3.tif'
pic = Image.open(file)
pic.size
print np.array(pic.getdata()).reshape(512,512)
print "These are the correct values"

pic2 = libtiff.TIFF.open(file).read_image()
print pic2

print "Same value, good values"
pic_pims = pims.TiffStack(file)
print pic_pims
print pic_pims[0]
print pic_pims[1]
print pic_pims[2]
print pic_pims[3]
print "not the good tif values"
```

And here is the output!
```
$ python pimsproblem.py 
[[692 691 695 ..., 703 701 696]
 [699 698 702 ..., 711 708 704]
 [705 703 707 ..., 716 713 709]
 ..., 
 [682 682 686 ..., 693 690 686]
 [680 679 683 ..., 690 688 684]
 [684 683 687 ..., 695 692 688]]
These are the correct values
[[692 691 695 ..., 703 701 696]
 [699 698 702 ..., 711 708 704]
 [705 703 707 ..., 716 713 709]
 ..., 
 [682 682 686 ..., 693 690 686]
 [680 679 683 ..., 690 688 684]
 [684 683 687 ..., 695 692 688]]
Same value, good values
<Frames>
Source: TRIMcapture_MG132_06_R3D_D3D_3.tif
Length: 34 frames
Frame Shape: 512 x 512
Pixel Datatype: uint16
[[46082 45826 46850 ..., 48898 48386 47106]
 [47874 47618 48642 ..., 50946 50178 49154]
 [49410 48898 49922 ..., 52226 51458 50434]
 ..., 
 [43522 43522 44546 ..., 46338 45570 44546]
 [43010 42754 43778 ..., 45570 45058 44034]
 [44034 43778 44802 ..., 46850 46082 45058]]
[...]

PR #145, opened 2015-04-21
FIX Fixes wrong use of ndarray.byteswap in TiffStack_libtiff

caspervdw:
This fixes https://github.com/soft-matter/pims/issues/141. The parameter in `ndarray.byteswap()` determines whether the byteswap is inplace. So the use was wrong.

Also, `ndarray.newbyteorder()` changes the dtype byteorder instead of physically swapping the data.

### DIFF
--- a/pims/tiff_stack.py
+++ b/pims/tiff_stack.py
@@ -219,7 +219,9 @@ class TiffStack_libtiff(FramesSequence):
         if j > self._count:
             raise ValueError("File does not contain this many frames")
         self._tiff.SetDirectory(j)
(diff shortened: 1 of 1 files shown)


tacaswell:
I am a bit confused here.  As it was does that mean the byteswap was _always_ happening and we only saw the error when it didn't need to be swapped?

caspervdw:
Exactly. `byteswap()` always swaps the byteorder, so we only see the bug if it is not needed. The `inplace` parameter has no effect as it is implemented now.

tacaswell:
Thanks, that was subtle bug.

PR #146, opened 2015-04-21
ImageSequence factory

danielballan:
This is useful for quickly building a subclass ImageSequence with a customized `imread` function.

### DIFF
--- a/pims/image_sequence.py
+++ b/pims/image_sequence.py
@@ -347,6 +347,12 @@ def customize_image_sequence(imread_func, name=None):
     Returns
     -------
     class : a subclass of ImageSequence
+
+    Example
+    -------
+    >>> # my_func accepts a filename and returns a numpy array
+    >>> MyImageSequence = customize_image_sequence(my_func)
+    >>> frames = MyImageSequence('path/to/my_weird_files*')
     """
     class CustomImageSequence(ImageSequence):
         def imread(self, filename, **kwargs):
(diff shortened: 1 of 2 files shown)


pims/image_sequence.py
```diff
@@ -332,3 +332,31 @@ def __repr__(self):
...
+    Parameters
+    ----------
+    imread_func : callable
+        image reader
+    name : str or None
+        name of class returned; if None, 'CustomImageSequence' is used.
+
+    Returns
+    -------
+    class : a subclass of ImageSequence
```
tacaswell: The return type is 'type'  and this needs a description body please.

tacaswell:
I like this :)

pims/image_sequence.py
```diff
@@ -332,3 +332,33 @@ def __repr__(self):
                                   dtype=self.pixel_type,
                                   C=self._sizeC,
                                   Z=self._sizeZ)
+
+
+def customize_image_sequence(imread_func, name=None):
+    """Class factory for ImageSequence with customized image reader.
+
```
licode: Neat. So users can write their own reader function, like image reader or hdf5 reader, and use it here. Similar to the idea of file store, or file store borrows some idea here?

danielballan: Right. Specifically, this works for readers where each file is a separate frame. Otherwise, one must right their own subclass of `FramesSequence`.

caspervdw:
Great! This is really helpful. I tested it and I had one issue: if `skimage` is installed the argument `plugin` will be passed to the `imread` function, which will give an exception.

I suggest removing the `plugin` argument from `ImageSequence` and instead use a general `**kwargs` argument, that is passed to imread altogether.

tacaswell:
:+1: on refactoring the plugin stuff.  That should probably be pulled out of the general case way, if you want to use a specific plug in do it via this factory + a closure over the plugin.

tacaswell:
Merging this, the refactor of the kwargs will come later.

PR #147, opened 2015-04-22
ENH: Slicing on a pims reader returns reusable, sliceable, sized objects.

danielballan:
PIMS readers have always supported random access, and slicing on them has always returned a generator, lazily loading the image data. Like all Python generators, this generator has no length, cannot be sliced on, and can only be traversed one time. Reminder example:

```
>>> images = ImageSequence(...)
>>> images[5]  # works
>>> list(images)  # works
>>> list(images)  # works again
>>> my_slice = images[:8]  # this is a generator
>>> my_slice[5]  # not allowed
>>> list(my_slice)  # works once, but not again
```

In this PR, slicing on a PIMS reader returns a new object. The objecct behaves more like the reader that spawned it. That is, `my_slice` itself becomes sliceable, and it can be traversed multiple times. It also knows its length.

I'll use a cute demo reader, using letters of the alphabet instead of actual images, to succinctly show the new behavior.

```
In [3]: class DemoReader(pims.ImageSequence):
    def imread(self, filename, **kwargs):
        return np.array([[filename]])
   ...:     

```

The original reader object is sliceable the same as it always has been.

```
In [4]: images = DemoReader(list('abcdefghij'))

In [5]: images[0]
Out[5]: 
Frame([['a']], 
      dtype='|S1')

In [6]: len(images)
Out[6]: 10

In [7]: list(images)
Out[7]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['b']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['d']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['f']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1'), Frame([['h']], 
       dtype='|S1'), Frame([['i']], 
       dtype='|S1'), Frame([['j']], 
       dtype='|S1')]

In [8]: list(images)
Out[8]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['b']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['d']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['f']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1'), Frame([['h']], 
       dtype='|S1'), Frame([['i']], 
       dtype='|S1'), Frame([['j']], 
       dtype='|S1')]
```

But now slicing on `images` returns an object that can be listed more than once...
```
In [9]: slice1 = images[:8]

In [10]: list(slice1)
Out[10]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['b']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['d']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['f']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1'), Frame([['h']], 
       dtype='|S1')]

In [11]: list(slice1)
Out[11]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['b']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['d']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['f']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1'), Frame([['h']], 
       dtype='|S1')]

```

...and is recursively sliceable! o_O

```
In [13]: slice2 = slice1[::2]

In [14]: list(slice2)
Out[14]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1')]
```

The iterators behind these various slices are independent. They do not interfere with each other. (Listing `images` still works....)

```
In [15]: list(images)
Out[15]: 
[Frame([['a']], 
       dtype='|S1'), Frame([['b']], 
       dtype='|S1'), Frame([['c']], 
       dtype='|S1'), Frame([['d']], 
       dtype='|S1'), Frame([['e']], 
       dtype='|S1'), Frame([['f']], 
       dtype='|S1'), Frame([['g']], 
       dtype='|S1'), Frame([['h']], 
       dtype='|S1'), Frame([['i']], 
       dtype='|S1'), Frame([['j']], 
       dtype='|S1')]
```

No image is data is copied. The only caching involved in this is the caching of indices (frame numbers).

The tests pass, but I would like to add more tests before we merge. This reaches into the heart of PIMS and does some dizzying things with iterators. Please give it a spin!

### DIFF
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -430,26 +430,34 @@ def _index_generator(new_indices, old_indices):
     """Find locations of new_indicies in the ref. frame of the old_indices.
     
     Example: (1, 3), (1, 3, 5, 10) -> (3, 10)
-    
+
     The point of all this trouble is that this is done lazily, returning
     a generator without actually looping through the inputs."""
     # Use iter() to be safe. On a generator, this returns an identical ref.
     new_indices = iter(new_indices)
-    old_indices = iter(old_indices)
     n = next(new_indices)
+    last_n = None
     done = False
-    for i, o in enumerate(old_indices):
-        if done:
-            raise StopIteration
-        if i == n:
-            try:
-                n = next(new_indices)
-            except StopIteration:
-                done = True
-                # Don't stop yet; we still have one last thing to yield.
-            yield o
-        else:
-            continue
+    while True:
+        old_indices_, old_indices = itertools.tee(iter(old_indices))
+        for i, o in enumerate(old_indices_):
+            # If new_indices is not strictly monotonically increasing, break
+            # and start again from the beginning of old_indices.
+            if last_n is not None and n <= last_n:
+                last_n = None
+                break
+            if done:
+                raise StopIteration
+            if i == n:
+                last_n = n
+                try:
+                    n = next(new_indices)
+                except StopIteration:
+                    done = True
+                    # Don't stop yet; we still have one last thing to yield.
+                yield o
+            else:
+                continue
 
 
 def pipeline(func):
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -225,7 +225,7 @@ class SliceableIterable(object):
                 # if we have a bool array, set up masking but defer
                 # the actual computation, returning another SliceableIterable
                 rel_indices = np.arange(len(self))[key]
-                indices = index_generator(rel_indices, abs_indices)
+                indices = _index_generator(rel_indices, abs_indices)
                 new_length = key.sum()
                 return SliceableIterable(self._ancestor, indices, new_length)
             if any(_k < -_len or _k >= _len for _k in key):
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -133,6 +133,101 @@ Pixel Datatype: {dtype}""".format(w=self.frame_shape[0],
                                   dtype=self.pixel_type)
 
 
+class SliceableIterable(object):
+
+    def __init__(self, ancestor, indices, length):
+        """A generator that support fancy indexing, returning another generator
+
+        Parameters
+        ----------
+        ancestor : object
+            must support __getitem__ with an integer argument
+        indices : iterable
+            giving indices into `ancestor`
+        length: integer
+            length of indicies
+            (must be given explicitly because indices may be a generator)
+
+        Examples
+        --------
+        # Slicing on a SliceableIterable returns another SliceableIterable...
+        >>> v = SliceableIterable([0, 1, 2, 3], range(4), 4)
+        >>> v1 = v[:2]
+        >>> type(v[:2])
+        SliceableIterable
+        >>> v2 = v[::2]
+        >>> type(v2)
+        SliceableIterable
+        >>> v2[0]
+        0
+        # ...unless the slice itself has an unknown length, which makes
+        # slicing impossible.
+        >>> v3 = v2((i for i in [0]))  # argument is a generator
+        >>> type(v3)
+        UnsliceableIterable
+        """
+        self._len = length
+        self._ancestor = ancestor
+        self._indices = indices
+        self._counter = 0
+
+    def __iter__(self):
+        indices, self._indices = itertools.tee(self._indices)
+        return (self._ancestor[i] for i in indices)
+
+    def __len__(self):
+        return self._len
+
+    def __getitem__(self, key):
+        """for data access"""
+        _len = len(self)
+        try:
+            # Advancing abs_indices won't affect this new copy of
+            # self._ancestor._indices.
+            abs_indices, self._ancestor._indices = itertools.tee(
+                    self._ancestor._indices)
+        except AttributeError:
+            abs_indices = range(len(self._ancestor))
+
+        if isinstance(key, slice):
+            # if input is a slice, return another SliceableIterable
+            rel_indices = range(*key.indices(_len))
+            indices = _index_generator(rel_indices, abs_indices)
+            start, stop, step = key.indices(_len)
+            new_length = (stop - start) // step
+            return SliceableIterable(self._ancestor, indices, new_length)
+        elif isinstance(key, collections.Iterable):
+            # if the input is an iterable, doing 'fancy' indexing
+
+            if isinstance(key, np.ndarray) and key.dtype == np.bool:
+                # if we have a bool array, set up masking but defer
+                # the actual computation, returning another SliceableIterable
+                rel_indices = np.arange(len(self))[key]
+                indices = index_generator(rel_indices, abs_indices)
+                new_length = key.sum()
+                return SliceableIterable(self._ancestor, indices, new_length)
+            if any(_k < -_len or _k >= _len for _k in key):
+                raise IndexError("Keys out of range")
+            try:
+                new_length = len(self._indices)
+            except TypeError:
+                # The key is a generator; return a plain old generator.
+                # Without knowing the length, we can't give a SliceableIterable.
+                gen = (self[_k if _k >= 0 else _len + _k] for _k in key)
+                return gen
+            else:
+                # The key is a list of in-range values. Build another
+                # SliceableIterable, again deferring computation.
+                rel_indices = ((_k if _k >= 0 else _len + _k) for _k in key)
+                indices = _index_generator(rel_indices, abs_indices)
+                return SliceableIterable(self._ancestor, indices, _len)
+        else:
+            if key < -_len or key >= _len:
+                raise IndexError("Key out of range")
+
+            return self._ancestor[key if key >= 0 else _len + key]
+
+
 class FramesSequence(FramesStream):
     """Baseclass for wrapping data buckets that have random access.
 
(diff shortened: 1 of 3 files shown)


nkeim:
From a purely functional viewpoint, this is great. I think that "kinda like a lazily-loading movie-aware ndarray" is a great model for pims.

This addresses an actual use case I was considering for pims: when we take data, we just stream from the camera into one monolithic movie file, then later split it into individual experiments based on gaps in the frame timestamps. The way our experiment is set up, this relieves the experimenter of a surprising amount of tedium. So being able to slice a large movie like this could save multiple processing steps and data copying.

tacaswell:
Talking to @danielballan about this in the car ride to the train station we should also add (mostly just here for discussion purposes, not to do in this PR)

```python
p = PimsPims(existing_pims_object, index_remapping_array)
p = pims.concat(pim1, pim2. pim3, ...)
```

The first just maps `p[n]` -> `existing_pims_object[index_remapping_array[n]]` and the `concat` return an object lets you pretend multiple pims objects are the same movie (I had something like this in grad school to stitch together multiple partially successful analysis runs).

danielballan:
```
p = PimsPims(existing_pims_object, index_remapping_array)
```

can be accomplished by

```
from pims import SliceableIterator
p = SliceableIterator(existing_pims_object, index_mapping_array)
```

And if your mapping is a generator, provide its length as the third (optional) parameter.

pims/base_frames.py
```diff
@@ -133,45 +133,145 @@ def __repr__(self):
...
         elif isinstance(key, collections.Iterable):
             # if the input is an iterable, doing 'fancy' indexing
 
             if isinstance(key, np.ndarray) and key.dtype == np.bool:
-                # if we have a bool array, do the right thing
-                return (self.get_frame(_k) for _k in np.arange(len(self))[key])
+                # if we have a bool array, set up masking but defer
+                # the actual computation, returning another SliceableIterable
+                rel_indices = np.arange(len(self))[key]
+                indices = index_generator(rel_indices, abs_indices)
```
caspervdw: This should be `_index_generator`

pims/base_frames.py
```diff
@@ -133,45 +133,145 @@ def __repr__(self):
                                   dtype=self.pixel_type)
 
 
-class FramesSequence(FramesStream):
-    """Baseclass for wrapping data buckets that have random access.
+class SliceableIterable(object):
```
caspervdw: The class name doesn't match with the import in api.py. I think renaming this to `SliceableIterator` is the best option, this also goes to multiple instances below.

caspervdw:
Very nice! This also allows block-reading several images at once with `np.array(frames[:5])`.

Just one thought, with the `Bioformats` reader there is a `close()` method. The following will lead to an exception:

    slicedframe = frames[5:]
    slicedframe.close()
    frames[0]

It will be more user-friendly to filter out calls to `close()` by giving `SliceableIterator` an own method `close()` which does nothing (or prints a warning that the ancestor object is not closed).

pims/base_frames.py
danielballan: Oops. I changed api.py, actually. In the course of writing this class, I learned that iterable and iterator have [specific meanings](http://stackoverflow.com/a/9884259/1221924), and this is an iterable.

danielballan:
@caspervdw Thanks for the heads-up about `close()`. If you don't think `slicedframe` should be able to close its "ancestor" `frames`, then I think the best option is to have `slicedframe.close()` raise an error, leaving no room for confusion: the file is still open. Sound reasonable?

tacaswell:
It cuts both ways though, if you close the ancestor the 'child' slice objects will stop working.  We probably should have some sort of weak-ref registry of child slicers on the ancestor which if there are any still around prevents closing (and calling close on child slicers removes them selves from that registry and calls close on the ancestor.  Which is starting to sound really complicated.

nkeim:
Yes. My own lesson from experience is "close a file in the same code block where you opened it, or don't close it at all." Pretty much any other practice generates complications and/or surprises. So I'm in favor of @danielballan 's idea to raise when `close()` is called on a child, which would effectively prevent file operations from being performed by code that doesn't know it's been passed a child. This would prevent a large class of problems that are otherwise difficult to debug, especially in the vein of "how can this file still be open?!".

The scenario @tacaswell presents would really stink, but it's probably easier to debug because an exception is raised the first time the user tries to get a frame from a child. And it can be avoided by never calling `close()` in the first place, which seems like it would be OK in the vast majority of use cases — the files are opened read-only, after all. So it seems like a solution is less important.

Can the Bioformats reader work as a context manager? That would seem like The Right Way To Do It.

tacaswell:
The context manager stuff _should_ all be in place and 'just work'

caspervdw:
But using context manager with a slicedframe will call close() in the end, raising an exception with Dans proposal.

Either way, I am in favour of a close() routine that either raises an exception or prints a warning message and then does nothing.

There is also a __del__() routine in bioformats that again calls close, so ancestors will be closed at some point.

tacaswell:
I think the idea of the context manager is that any sub-slicing you do is
done _inside_ the context manager (and the functions you call inside of
it).  My understanding was that although the context object is still in the
same space after the context block exits, using it outside of the context
manager block is at your own risk.

caspervdw:
Ah I see thanks for the explanation.

danielballan:
The latest commit adds a no-op `close` method. It sounds like we agree that leaving a read-only file open is not a big deal, while accidentally closing a parent via its child could be frustrating.

I still want to add tests before we merge.

tacaswell:
Did anyone ever set up ASV on pims?

danielballan:
Not that I know of. There is no pims-bench.

danielballan:
I added a slew of tests, including testing out the fourth level of nested slicing (a slice of a slice of a slice of a slice). I am confident that this does not break existing slicing behavior.

danielballan:
Scratch that. This is maddeningly subtle. There is more work to be done here.

danielballan:
@nkeim @caspervdw @tacaswell I added another simple but very useful enhancement to this PR, revealing my motivation for this slicing work.

With a simple decorator, you can apply any image-processing function to a pims object and get lazy evaluation and slicing.

```
@pims.pipeline
def color_channel(image, channel):
    return image[channel, :, :]

red = color_channel(v, 0)
green = color_channel(v, 1)
```

Now `red` and `green` are `SliceableIterator` objects. They close over the function arguments, lazily processing the images on the way out.

Pipeline functions are also composable: `func1(func2(func3(v))))` does the right thing.

The "simple case" behavior of `color_channel` is unchanged. If you pass it a numpy array or a `pims.Frame`, the pipeline decorator does nothing.

danielballan:
Bump. I think this is ready to merge. The tests are solid. Someone should look over the added pipeline code, but it's short and straightforward.

pims/base_frames.py
```diff
@@ -133,45 +134,159 @@ def __repr__(self):
...
+
+    def __init__(self, ancestor, indices, length=None):
+        """A generator that support fancy indexing, returning another generator
```
caspervdw: supports

caspervdw:
I left a very picky comment, for the rest I think it is ready to merge.

pims/base_frames.py
```diff
@@ -133,45 +134,163 @@ def __repr__(self):
...
+class SliceableIterable(object):
+
+    def __init__(self, ancestor, indices, length=None):
+        """A generator that supports fancy indexing
+
+        When sliced using any iterable with a known length, it return another
+        object like itself, a SliceableIterable. When sliced with an integer,
+        it returns the data payload.
+
+        Also, this retains the attributes of the "ancestor" object that
```
tacaswell: This is not quite true, you keep passing the base ancestor through, not it's immediate parent.

pims/base_frames.py
```diff
@@ -133,45 +134,163 @@ def __repr__(self):
...
+        created it.
+
+        Parameters
+        ----------
+        ancestor : object
+            must support __getitem__ with an integer argument
+        indices : iterable
+            giving indices into `ancestor`
+        length: integer
```
tacaswell: missing `, optional`

pims/base_frames.py
```diff
@@ -133,45 +134,163 @@ def __repr__(self):
...
+            length of indicies
+            This must be given explicitly if indices is a generator,
```
tacaswell: This needs to be a bit clearer due to the fall-back-to-a-generator behavior.

tacaswell: nope, I am confused :sheep:

pims/base_frames.py
```diff
@@ -306,3 +425,99 @@ def __repr__(self):
                                   h=self.frame_shape[1],
                                   count = len(self),
                                   dtype=self.pixel_type)
+
+
+def _index_generator(new_indices, old_indices):
+    """Find locations of new_indicies in the ref. frame of the old_indices.
```
tacaswell: Add a note that new indicies must be strictly monotonic ?

pims/tests/test_common.py
```diff
@@ -117,6 +120,120 @@ def invert(image):
...
+        compare_slice_to_list(slice4, list('hi'))
+
+        # Give me another!
+        slice1 = self.v[2:]
+        compare_slice_to_list(slice1, list('cdefghij'))
+        slice2 = slice1[0::2]
+        compare_slice_to_list(slice2, list('cegi'))
+        slice3 = slice2[:]
+        compare_slice_to_list(slice3, list('cegi'))
+        slice4 = slice3[:-1]
```
tacaswell: can you add a test of `slice3[::-1]` ?

pims/base_frames.py
danielballan: With new commits, indices do not need to be monotonic. Tests confirm this.

tacaswell:
:beers:

Issue #150, opened 2015-05-06
Roadmap for PIMS


danielballan:
Here are my thoughts on the near future of PIMS development.

Before tagging v0.3:
- Add a few more readers with associated optional dependencies.
- Remove the janky `plugin` argument from `ImageSequence`, and instead create separate classes (`ImageSequence_skimage`, `ImageSequence_mpl`, `ImageSequence_scipy`) the way we do for the variants of `TiffStack`. As with `TiffStack`, there will be a convenient `ImageSequence` alias pointing at the most reliable variant available.
- Merge #147 with recursive slicing and the new "pipeline" idea.

Then, once v0.3 is tagged, I'd like to make a more serious break change. If the pipelines are working well, I'd like to simplify all the built-in readers by removing `process_func`, `as_grey`, `dtype` as standard arguments. Any of those tasks -- and more -- can be accomplished more transparently and more generally by "pipeline" functions. I'd prefer the reader classes to be as lightweight and simple as possible. (Note that users could still implement their own readers that have a `process_func` arg or whatever they want; I'm just talking about the built-in readers here.)

With that major change in place, then I think it's time to tag 1.0.

Thoughts?

danielballan:
@nkeim if you have time to read over this plan I'd like your input.

caspervdw:
I think I am not long enough into pims to really comment on when to tag a release. But I would additionally like more homogeneous support for multidimensional files (Z, colorchannels, RGB channels). There are already 2 3D readers (`ImageSequence3D` and `Bioformats`) but it should be put in a more general wrapping.

Maybe something base_classes `FramesSequence3D`, `FramesSequenceMultichannel`, `FramesSequence3DMultichannel`.

This by the way is already being discussed inside https://github.com/soft-matter/pims/issues/124

danielballan:
@caspervdw My goal with 1.0 is to correct API choices that were in retrospect not good and to commit to not breaking things for awhile. This does not preclude adding new functionality. Whether or not we get the desired 3D and color support into 1.0, I am open to issuing a 1.1 as soon as necessary.

What do you envision `FramesSequence3D` doing? I think of `FrameSequence` as N-dimensional, leaving to the the readers to do whatever they need to do in `__init__` and `get_frame`. Should we add something to `FramesSequence` to make that easier? I can see that the idea of "loop order" is important and maybe should work its way up. It's just not clear to me whether it should be a base class method or a shared utility function or something else.

nkeim:
[...]

Issue #124, opened 2014-12-20
Order of dimensions in multichannel images


caspervdw:
In PR https://github.com/soft-matter/pims/pull/118 and https://github.com/soft-matter/pims/pull/109, I started working on importing images with more than one channel (C), focal plane (Z) and time points (T). The dimensional order I've adopted is (TCZYX), starting with the outer dimension.

I've read https://github.com/scikit-image/scikit-image/pull/1280 and the scikit convention is to put multichannels in the last dimension. I have a strong tendency to go along with this convention, but it does raise a few issues:

- this dimensional order is unconventional in microscopy (except for RGB). For example, ImageJ, which is the most standard image analysis tools always has YX as last two dimensions.
- `trackpy` operates on grayscale images, so a conversion `process_func` is always necessary if you import a multichannel. For instance, subtracting two channels is less computer-intensive when channel is the first dimension. (compare `image[0] - image[1]` to `image[:,:,0] - image[:,:,1]`)
- `_repr_html_` in `frame.py` distinguishes multichannels from RGBs based on the location of multichannel axis. A 3-channel multichannel would be indistinguishable from RGB, unless we decide to give the `frame` object an `isRGB` property. Which I am in favour of.

Question is: do we keep it as it is or should we go along with scikit-image and put channels in the inner dimension? What is your opinion?

caspervdw:
PR 109 and PR 118 are sort of waiting for this decision. What do you think about this @tacaswell @danielballan?

danielballan:
Hey, sorry for the delay, first week working at BNL with @tacaswell! Reviewing now....

tacaswell:
Sorry, @danielballan has been busy defending and moving in the past month, I don't have as good of an excuse.

On further consideration, the skimage conventions are for storing bunches of images together in memory, where as pims is more about getting images stored in files into memory with out having to special-case file formats everywhere.  The file formats that drove Dan and I to work on this were derived from what we had to deal with, which were single streams of gray-scale or RGB frames, which leaks into the assumptions made by pims.

To that end, I think we can be flexible about this sort of ordering.  The sub-classes which _know_ about channels should be able to be configured at runtime as to the order they loop over or aggregate the raw frames.  Maybe something like:

```python
class ChannelZAware(self, ...., loop_order=['z', 'c', 't']):
     pass
```

so when you do `for im in cz_awar` it loops over all z at fixed (c, t) etc.  I don't claim this is the best scheme for notating this.

[...]

PR #151, opened 2015-05-18
ENH Multidimensional image base class

caspervdw:
This implements a base class that allows flexible reading and iteration through a file with multiple dimensions.  The properties `aggregate` and `iterate` define how each dimension functions, as discussed in #124 and #150 

I based the `ImageSequenceND` reader on this, I'll wait for your response until rebasing the `Bioformats` reader.

The base class itself is quite extensive, so that the derived classed can be slim:

    >>> class MDummy(Multidimensional):
    ...    @property
    ...    def pixel_type(self):
    ...        return 'uint8'
    ...    @property
    ...    def frame_shape_2D(self):
    ...        return self._frame_shape_2D
    ...    def __init__(self, shape, **dims):
    ...        for name in dims:
    ...            self.add_dim(name, dims[name])
    ...        self._frame_shape_2D = shape
    ...    def get_frame_2D(self, **ind):
    ...        return np.zeros(self.frame_shape_2D, dtype=self.pixel_type)

    >>> frames = MDummy((64, 64), t=80, c=2, z=10, m=5)
    >>> frames.aggregate = 'cz'
    >>> frames.iterate = 't'
    >>> frames.m.default = 3
    >>> frames[5]  # returns Frame at T=5, M=3 with shape (2, 10, 64, 64)

### DIFF
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -633,7 +633,7 @@ class FramesSequenceND(FramesSequence):
         as the order of `bundle_axes`.
         The last two elements have to be ['y', 'x'].
         """
-        return self._aggregate
+        return self._bundle_axes
 
     @bundle_axes.setter
     def bundle_axes(self, value):
@@ -655,7 +655,7 @@ class FramesSequenceND(FramesSequence):
         """ This determines which axes will be iterated over by the
         FramesSequence. The last element will iterate fastest.
         x and y are not allowed. """
-        return self._iterate_axes
+        return self._iter_axes
 
     @iter_axes.setter
     def iter_axes(self, value):
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -528,3 +528,57 @@ def pipeline(func):
                        "other objects, its behavior is "
                        "unchanged.\n\n") + process.__doc__
     return process
+
+
+class Multidimensional(FramesSequence):
+    @abstractproperty
+    def sizes(self):
+        pass
+
+    @abstractproperty
+    def dim_default(self):
+        pass
+
+    @abstractproperty
+    def dim_names(self):
+        pass
+
+    @abstractproperty
+    def dim_aggregate(self):
+        pass
+
+    @abstractproperty
+    def dim_loop_over(self):
+        pass
+
+    def __len__(self):
+        return np.prod([self.sizes[d] for d in self.dim_loop_over])
+
+    def frame_shape(self):
+        shape = [self.sizes[d] for d in self.dim_aggregate]
+        return shape + [self.sizes['Y'], self.sizes['X']]
+
+    def get_frame(self, i):
+        ind = {n: self.dim_default[n] for n in self.dim_names}
+        i_prev = 0
+        for n, dim in enumerate(self.dim_loop_over):
+            size = int(np.prod([self.sizes[d] for d in self.dim_loop_over[n+1:]]))
+            ind[dim] = (i - i_prev) // size
+            i_prev += ind[dim] * size
+
+        Nframes = int(np.prod([self.sizes[d] for d in self.dim_aggregate]))
+
+        result = np.empty((Nframes, self.sizes['Y'], self.sizes['X']))
+        for n in range(Nframes):
+            result[n] = self.get_frame_2D(**ind)
+            for d in self.dim_aggregate:
+                ind[d] += 1
+                if ind[d] >= self.sizes[d]:
+                    ind[d] = 0
+                else:
+                    break
+        return Frame(result.reshape(self.frame_shape()))
+        
+    @abstractmethod  
+    def get_frame_2D(self, **ind):
+        pass
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -530,88 +530,7 @@ def pipeline(func):
     return process
 
 
-class FramesDimension(object):
-    """
-    This class defines a dimension for use in the Multidimensional base class.
-
-    Parameters
-    ----------
-    name : string, dimension name
-    size : int, size of dimension
-    aggregate : boolean, default False
-        when True, this sets the dimension to be aggregated inside the Frame
-    iterate : boolean, default True
-        when True, this sets the dimension to be iterable
-    default : int
-        when aggregate and iterate are both False, this is the fallback value
-    """
-    def __init__(self, name, size, aggregate=False, iterate=True, default=0):
-        if name.lower() in ['x', 'y']:
-            raise ValueError('The names x and y are reserved and cannot be '
-                             'used.')
-        self.name = name
-        self._default = default
-        self.size = size
-        if aggregate and iterate:
-            raise ValueError('Dimensions cannot aggregate and be iterable '
-                             'simultaneously.')
-        self._aggregate = aggregate
-        self._iterate = iterate
-
-    @property
-    def size(self):
-        return self._size
-
-    @size.setter
-    def size(self, value):
-        if value <= 0:
-            raise ValueError('Dimension size should be greater than zero.')
-        self._size = int(value)
-        self.default = self._default  # recheck default value validity
-
-    @property
-    def default(self):
-        return self._default
-
-    @default.setter
-    def default(self, value):
-        if value >= self.size:
-            raise ValueError('Default value out of bounds')
-        self._default = value
-
-    @property
-    def aggregate(self):
-        return self._aggregate
-
-    @aggregate.setter
-    def aggregate(self, value):
-        self._aggregate = value
-        if value:  # iterate and aggregate cannot both be True
-            self._iterate = False
-
-    @property
-    def iterate(self):
-        return self._iterate
-
-    @iterate.setter
-    def iterate(self, value):
-        self._iterate = value
-        if value:  # iterate and aggregate cannot both be True
-            self._aggregate = False
-
-    def __repr__(self):
-        s = "<FramesDimension>; Name: {0}; Size: {1}; Status: "
-        s = s.format(self.name, self.size)
-        if self.iterate:
-            s += "iterate"
-        elif self.aggregate:
-            s += "aggregate"
-        else:
-            s += "defaults to {}".format(self.default)
-        return s
-
-
-class Multidimensional(FramesSequence):
+class FramesSequenceND(FramesSequence):
     """ A base class defining a FramesSequence with an arbitrary number of
     dimensions. The properties `aggregate` and `iterate` define the functions
     of each dimension.
@@ -641,16 +560,20 @@ class Multidimensional(FramesSequence):
     >>> frames.dims['m'].default = 3
     >>> frames[5]  # returns Frame at T=5, M=3 with shape (2, 10, 64, 64)
     """
-    def clear_dims(self):
-        self._dims = []
+    def _dim_init(self):
+        self._sizes = {}
+        self._default_coords = {}
+        self._iterate = []
+        self._aggregate = []
 
-    def add_dim(self, name, size, aggregate=False, iterate=True,
-                default=0):
-        new_dim = FramesDimension(name, size, aggregate, iterate, default)
-        self._dims += [new_dim]
+    def _dim_add(self, name, size, default=0):
+        if not hasattr(self, '_sizes'):
+            self._dim_init()
+        self._sizes[name] = size
+        self._default_coords[name] = default
 
     def __len__(self):
-        return np.prod([d.size for d in self._dims if d.iterate])
+        return np.prod([self._sizes[d] for d in self._iterate])
 
     @abstractproperty
     def frame_shape_2D(self, **ind):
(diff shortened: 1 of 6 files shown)


tacaswell:
Awesome.  I don't have time to read the code in detail right now, but it looks like you have sorted out how to solve some problems that I looked and gave up on as too hard :smile: .

pims/base_frames.py
```diff
@@ -528,3 +528,267 @@ def process(img_or_iterable, *args, **kwargs):
...
+    @property
+    def sizes(self):
+        """ Returns a dict of all dimension sizes, including x and y. """
+        result = {d.name: d.size for d in self._dims}
+        result['y'], result['x'] = self.frame_shape_2D
+        return result
+
+    @property
+    def aggregate(self):
+        """ This determines which dimensions will be aggregated into one.
```
danielballan: typo

danielballan:
Yes, this is good stuff. I will review carefully this week. I wonder if the multidimensional base class should become the only base class, making all of this stuff available by default.

pims/base_frames.py
```diff
@@ -528,3 +528,267 @@ def process(img_or_iterable, *args, **kwargs):
...
+        """ Sets an empty value for self._dims without using __init__.
+
+        Enables dimensions to be accessed by for instance frames.c.
+        Existing methods always precede over this: a dimension named 'ndim'
+        would not be accessible by frames.ndim. """
+        if key == '_dims':
+            return []
+        for dim in self._dims:
+            if key == dim.name:
+                return dim
```
danielballan: I think this approach fails silently if I try to access a dim that does not exist.

caspervdw: That's right, is there a proper way to solve this? Just raise an `AttributeError`?

tacaswell: Yes, that will let python fall back to `__getattribute__` and then if that fails bubble up to the user.

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
     return process
+
+
+class FramesDimension(object):
+    """
+    This class defines a dimension for use in the Multidimensional base class.
+
+    Parameters
+    ----------
+    name : string, dimension name
```
tacaswell: Please put description on new line other wise the numpydo parser will identify name as having the type 'string, dimension name'.

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+    size : int, size of dimension
+    aggregate : boolean, default False
```
tacaswell: Things that are optional should be `name : type, optional`  with the default given in the docstring.

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
                        "other objects, its behavior is "
                        "unchanged.\n\n") + process.__doc__
     return process
+
+
+class FramesDimension(object):
```
tacaswell: Might be worth looking at http://boltons.readthedocs.org/en/latest/namedutils.html#boltons.namedutils.namedlist which is mutable `namedtuple`. It will be lighter weight at runtime and less code for use to maintain.

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+    ...    def get_frame_2D(self, **ind):
+    ...        return np.zeros(self.frame_shape_2D, dtype=self.pixel_type)
+
+    >>> frames = MDummy((64, 64), t=80, c=2, z=10, m=5)
+    >>> frames.aggregate = 'cz'
+    >>> frames.iterate = 't'
+    >>> frames.dims['m'].default = 3
+    >>> frames[5]  # returns Frame at T=5, M=3 with shape (2, 10, 64, 64)
+    """
+    def clear_dims(self):
```
tacaswell: Is this and `add_dim` really useful as public functions?  I have not mad it farther down the code yet, but it seems that once the reader is created, the dimensions should not change.

pims/base_frames.py
tacaswell: The other thing that occurs to me is we should be aware of https://github.com/xray/xray and see what we can steal from them.

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+    def iterate(self):
+        """ This determines which dimensions will be iterated over by the
+        FramesSequence. The last element will iterate fastest. """
+        return [d.name for d in self._dims if d.iterate]
+
+    @iterate.setter
+    def iterate(self, value):
+        for dim_iterate in value:
+            if dim_iterate.lower() in ['x', 'y']:
+                raise ValueError('Iterate does not take dimensions x or y.')
```
tacaswell: Induldging in a bit of bike shedding:

Why not `RuntimeError`?

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+            if dim_iterate not in self.dims:
+                raise ValueError(("Dimension named '{}' does not exist " +
```
tacaswell: bikeshedding: why not `KeyError`

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+                                  "in this image.").format(dim_iterate))
+
+        new_dims_iter = [0] * len(value)
```
tacaswell: bikeshedding : use `None` for the place holder

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+    def get_frame_2D(self, **ind):
+        """ The actual frame reader, defined by the subclassed reader.
+
+        This method should take exactly one keyword argument per dimension,
+        reflecting the index along each dimension. It returns a two dimensional
+        ndarray with shape `frame_shape_2D` and dtype `pixel_type`.
+        """
+        pass
+
+    def get_frame(self, i):
```
tacaswell: This is subtle as all get out

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+    name : string, dimension name
+    size : int, size of dimension
+    aggregate : boolean, default False
+        when True, this sets the dimension to be aggregated inside the Frame
+    iterate : boolean, default True
+        when True, this sets the dimension to be iterable
+    default : int
+        when aggregate and iterate are both False, this is the fallback value
+    """
+    def __init__(self, name, size, aggregate=False, iterate=True, default=0):
```
tacaswell: can you add the attribute `i` to this class?

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+        pass
+
+    def get_frame(self, i):
+        """ Returns a Frame of shape deterimend by aggregate. The property
+        iterate (together with the dimension default index) determines which.
+        multidimensional index is returned. """
+        # identify the indices to take along each dimension
+        i_prev = 0
+        for n, dim in enumerate(self._dims):
+            dim.i = dim.default
```
tacaswell: won't this do funny things if the default is set to 1/True?  My understanding from reading above was that this value was the default values for iterate/aggregate?

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+        ndarray with shape `frame_shape_2D` and dtype `pixel_type`.
+        """
+        pass
+
+    def get_frame(self, i):
+        """ Returns a Frame of shape deterimend by aggregate. The property
+        iterate (together with the dimension default index) determines which.
+        multidimensional index is returned. """
+        # identify the indices to take along each dimension
+        i_prev = 0
```
tacaswell: I am not super wild about using an attribute on the dimensions to keep track of where you are, I think this code would be easier to understand as something like

```python
sizes = np.array([d.size for d in self.dims if d.iterate], dtype='int')
cumsizes = np.cumprod(sizes)
current_index = np.zeros(len(self._dims), dtype=int)
```

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+                i_prev += dim.i * size
+
+        # initialize a stack of 2D arrays to collect the Frame
+        Nframes = int(np.prod(self.frame_shape[:-2]))
+        result = np.empty([Nframes] + list(self.frame_shape_2D),
+                          dtype=self.pixel_type)
+
+        # read all 2D frames
+        for n in range(Nframes):
+            result[n] = self.get_frame_2D(**{d.name: d.i for d in self._dims})
```
tacaswell: following the above:

```python
result[n] = self.get_frame_2d(**{d.name: indx for d, indx in zip(self._dims, cur_index)})
```

pims/base_frames.py
tacaswell: these have been dealt with :heavy_check_mark:

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+        return Frame(result, frame_no=i)
+
+    def __getattr__(self, key):
+        """ Sets an empty value for self._dims without using __init__.
+
+        Enables dimensions to be accessed by for instance frames.c.
+        Existing methods always precede over this: a dimension named 'ndim'
+        would not be accessible by frames.ndim. """
+        if key == '_dims':
+            return []
```
tacaswell: Is this to hide `_dims` or sholud this have been `return list(self._dims)`

pims/base_frames.py
```diff
@@ -528,3 +528,268 @@ def process(img_or_iterable, *args, **kwargs):
...
+        if self.iterate:
+            s += "iterate"
+        elif self.aggregate:
+            s += "aggregate"
+        else:
+            s += "defaults to {}".format(self.default)
+        return s
+
+
+class Multidimensional(FramesSequence):
```
tacaswell: This should probably have an 

```python

def __init__(self, *args, **kwargs):
    self._dims = []
    super(Multidimensional, self).__init__(self, *arg, **kwargs)
```

pims/image_sequence.py
```diff
@@ -248,87 +246,84 @@ class ImageSequence3D(ImageSequence):
...
     """
     def __init__(self, path_spec, process_func=None, dtype=None,
-                 as_grey=False, plugin=None, tzc_identifiers=None):
-        self.tzc_identifiers = tzc_identifiers
-        super(ImageSequence3D, self).__init__(path_spec, process_func,
+                 as_grey=False, plugin=None, dim_identifiers='tzc'):
+        if as_grey:
+            raise ValueError('As grey not supported for ND images')
+        self.dim_identifiers = dim_identifiers
+        super(ImageSequenceND, self).__init__(path_spec, process_func,
```
tacaswell: do the super first if you can.

pims/image_sequence.py
```diff
@@ -248,87 +246,84 @@ class ImageSequence3D(ImageSequence):
...
                                               dtype, as_grey, plugin)
+        if 't' in self.dims:
+            self.iterate = 't'  # iterate over t
+        if 'z' in self.dims:
+            self.aggregate = 'z'  # return z-stacks
 
     def _get_files(self, path_spec):
-        super(ImageSequence3D, self)._get_files(path_spec)
-        self._toc = np.array([filename_to_tzc(f, self.tzc_identifiers) \
+        super(ImageSequenceND, self)._get_files(path_spec)
```
tacaswell: I have been working on a python3 only project, the better super support is great (`super().foo` works as expected)

tacaswell:
Left a whole bunch of comments, over all :+1: 

any comments where I used the phrase 'bike shedding' feel free to give minimal thought and dismiss.

Sorry this has sat for so long

caspervdw:
Thanks for the thorough review, I hope to give this a round sometime this
week.
Op 19 jun. 2015 19:04 schreef "Thomas A Caswell" <notifications@github.com>:

pims/base_frames.py
caspervdw: I guess I took `ValueError` because the user is trying to set a non-existing dimension to `iterate`

pims/base_frames.py
caspervdw: Don't see exactly what you mean here, but I revised the code strongly so please take a look again at this part

pims/base_frames.py
caspervdw: Agreed, but this is a quite tricky part of the code. I now have this:

        # list sizes of iterate dimensions
        iter_sizes = [self._sizes[k] for k in self._iterate]
        # list how much i has to increase to get an increase of coordinate n
        iter_cumsizes = np.append(np.cumprod(iter_sizes[::-1])[-2::-1], 1)
        # calculate the coordinates and update the coords dictionary
        iter_coords = (i // iter_cumsizes) % iter_sizes

pims/base_frames.py
tacaswell: That was just a comment that this is a subtle problem and I think you got it more-or-less right (modulo where state is stored which is an implementation detail).

pims/base_frames.py
caspervdw: This was to properly initialize the _dims attribute without using __init__. But I moved it now to a check in `_dim_add`

pims/base_frames.py
caspervdw: Ok, may be best solution

pims/image_sequence.py
caspervdw: Can't, the `__init__` requires the `dim_identifiers`

caspervdw:
I refactored the code, tests pass (locally at least).

Inspired by xray I used the terminology _coordinate_ for an index along a specific dimension and _index_ for the value passed to `get_frame`. The transformation index --> coordinates according to `iterate` gave me some headaches and is (as far as I could see) not implemented in xray.

I kicked out the `FramesDimension` object and I just use dictionaries `_sizes` and `_default_coords` and lists `_aggrate` and `_iterate` to keep track of all the dimension properties. For clarity, the default coordinate is the coordinate that is used if the dimension is both not in `aggregate` and `iterate`

I didn't want to use the `__init__` because if you subclass for instance `ImageSequence` and `FramesSequenceND` together, you want to keep both `__init__` routines. Therefore I thought it to be more elegant to solve the dimension initialization in `_dim_add` if `_sizes` does not exist.

@tacaswell  I hope I covered all your suggestions, it's ready for another round of review and then I'll rebase and also base the Bioformats reader on this.

danielballan:
Slick. Thanks for tackling this subtle problem in a general way. Two comments:

1. I like the names borrowed from xray. The only name I don't like is aggregate, which to me suggested data reduction, as in SQL parlance or in pandas like: `df.groupby(...).agg(...)`. Are you open to using a different name? How about `bundle`?
2. Any `Frame.metadata` is ignored. This is the right choice for now -- better to drop it than propagate it badly -- but I just wanted to note it. Persisting metadata correctly and capturing more if it from the file formats we support can be a project for 0.3+.

caspervdw:
Rebased and added an extra update. Now `y` and `x` are treated equally as all dimensions. With some limitations due to the inherent 2D nature of image readers.

A thing left to do is handle interleaved formats (e.g. RGB)

caspervdw:
@danielballan I just implemented a metadata propagation support from `get_frame_2D` to `get_frame`. Metadata dicts of 2D frames get squashed in a dictionary of lists. If all values are the same in the list, it will be converted to a scalar. The derived reader may choose to include the coordinates of the 2D frames in the metadata.

I think there is no wrong metadata propagated in this way.

Separate thought: dictionaries aren't really the best format for storing multidimensional metadata.

danielballan:
OK, that sounds like a good approach to metadata. The general problem of propagating metadata through operations of Frame is much harder. (What happens when you add to Frames?) Pandas has raised this discussion and, for now, chosen to punt rather than guess and create some not-very-elegant conglomeration.

I like dictionaries for their flexible structure. What would you propose as an alternative? We should get this right going into 1.0.

Any comment on the nomenclature of "aggregate" versus an alternative like "bundle"? If you really like aggregate, I can be convinced, but I think both @tacaswell and I would prefer something that is not overloaded by existing tools.

caspervdw:
For me, "bundle" is ok. We can also do something more descriptive, such as "bundled_dims", or "bundled_axes".

What do you think of renaming "dim" to "axis", like in numpy?

As for the metadata, I guess the dictionaries are the best option for the general case. Just for the ND files, I would like to have some container that describes each plane (such as timestamp of each image), using multiindexing. A picture of size 10x2x512x512 should have a container of shape 10x2 that contains all the timestamps of each plane. Maybe an ndarray contained inside the dictionary is the best solution to that {'timestamps': ndarray}.

The way it is implemented now is as follows: {'z_index': list, 'c_index': list, 'timestamp': list}.

danielballan:
I like "axis." I usually go for verbose and descriptive, so bundled_axes
seems good to me. Bundled alone might confusingly suggest a Boolean.

I think an ndarray of metadata reflecting the structure of the bundled
Frame sounds right.

caspervdw:
Ok, I renamed the following:
* `iterate` -> `iter_axes`
* `aggregate` -> `bundle_axes`
* `dims` -> `axes`
* `_add_dim` -> `_init_axis`

Also I added support and tests for metadata propagation in ndarrays.

danielballan:
OK, after one more careful look from us is this ready to merge on your end,
Casper?

pims/image_sequence.py
```diff
@@ -248,90 +246,86 @@ class ImageSequence3D(ImageSequence):
...
+        This determines which dimensions will be iterated over by the
+        FramesSequence. The last element in will iterate fastest.
+        x and y are not allowed. Defaults to ['t'].
+    bundle_axes : iterable of strings
+        This determines which dimensions will be aggregated into one Frame.
+        The dimensions in the ndarray that is returned by get_frame has
+        the same order as the order in this list. The last two elements have
+        to be ['y', 'x']. Defaults to ['z', 'y', 'x'].
+    default_coords: dict of int
+        When a dimension is not present in both iterate and aggregate, the
```
danielballan: iterate and aggregate -> iter_axes and _bundle_axes

caspervdw:
It's ready for merge on my end, now also Bioformats is changed. Only thing is that the reader cannot return interleaved RGB or RGBA formats, because the last two axes are forced to ['y', 'x'].
But I don't have time to implement that until July 7th.

danielballan:
Looks like Travis is failing on bioformats:

```
AttributeError: 'BioformatsReader' object has no attribute '_source_dtype'
```

caspervdw:
Strange, I am sure tests passed locally... will have a look tomorrow!
Op 21 jun. 2015 10:11 schreef "Casper van der Wel" <caspervdw@gmail.com>:

caspervdw:
@danielballan @tacaswell I fixed the tests, all should be fine now. I'll be away until July 7th, so I hope it can be merged. Otherwise, feel free to edit the code.

danielballan:
Great. Thanks for pushing this through before your departure.

PR #158, opened 2015-06-24
Add reader for Norpix .seq files

nkeim:
This adds a simple reader class for Norpix's proprietary "sequence" format. This is the high-throughput format for [StreamPix software](https://www.norpix.com/products/streampix/streampix.php), which is a popular way to control high-performance cameras. The class reads image data and timestamps.

I based this code on my previously-published [standalone reader class](https://github.com/nkeim/py-norpix-reader), and on the cine reader already in pims. The reader currently supports uncompressed mono files with (in principle) arbitrary bit depth.

A ~1 MB sample file from StreamPix 6 is included for the unit tests. I also run tests against StreamPix 5 (a very slightly different format; sample file not included).

Support for `as_gray` and `process_func` is included, so in that sense, this could be merged now, or after #150 has run its course.

Writing this was surprisingly painless. Thank you, pims!


nkeim:
A corrected/improved/more fully vetted version of these changes was incorporated into #160.

Issue #159, opened 2015-06-24
ENH: Slicing-aware reader methods


nkeim:
This is an architectural issue at the intersection of #147, #153, and #158.

### The problem

Currently, pims `SliceableIterable` objects wisely forward unknown method calls to their ancestors, which allows methods to be "inherited" from the reader object that opened the file. However, those methods could take frame numbers as their arguments. For example, the Cine and (unmerged) NorpixSeq classes have methods for reading the timestamp of frame *i*. Being able to retrieve a timestamp without reading the entire image can make a large performance difference in some applications. In a similar fashion, these proprietary formats can store other (as yet unsupported) information such as voltages or event markers.

Therefore, methods like `get_time(frameno)` should be aware that they are being called via a child.

### A solution
`FramesSequence`-derived objects that want to support this feature provide the context `_use_map`, which would be used within the child `SliceableIterable` as follows (oversimplified for clarity):

```
# Call get_time() on the parent object
with self._ancestor.use_map(self.indices):
    self._ancestor.get_time(*args, **kw)
```

The reader class would then implement `get_time` as:

```
def get_time(self, frameno):
   actual_frameno = self._indexmap(frameno)
   # Return time of actual_frameno...
```

`use_map()`, `_indexmap()`, etc. would be implemented in the `FramesSequence` base class. `use_map()` could be thread-safe.

Thoughts? I'm open to more elegant solutions, or to the possibility that this is too much of a niche feature!

nkeim:
The idea I *just* posted is not going to work, because the `SliceableIterable` does not actually forward method calls; it just uses `__getattr__`.

The way to make it work would be to let the ancestor object supply the names of methods that need this mapping functionality.

nkeim:
To make sense of all this, I coded up the idea. You can [see the required changes here].(https://github.com/nkeim/pims/compare/seqreader...nkeim:sliced_methods?expand=1)

danielballan:
I had not considered this use case. I like your solution. I think we should merge your changes, update Cine, and consider moving the mapping stuff into FramesSequence proper sometime later.

caspervdw:
I think we could view this in the larger perspective of how to deal with metadata reading. Maybe a general Metadata object that collects al the get_something methods, correctly interprets the indices and also figures out the multidimensional case would be the most clear solution?

nkeim:
[...]

PR #160, opened 2015-06-25
ENH: Add Norpix, slicing-aware reader methods

nkeim:
This PR supersedes #158, and also addresses #159 using the general approach outlined there. With this PR, methods like `get_time(frameno)` behave the same way on slices of movies as they do on entire movies. This brings us closer to a world in which `FrameSequence` and `SliceableIterable` are interchangeable.

NorPix (see #158) and Cine classes have been updated to use this functionality. It would probably be best to avoid merging this PR before #158 has been approved and merged.


pims/base_frames.py
```diff
@@ -8,6 +8,9 @@
 import numpy as np
 import collections
 import itertools
+import functools
+import contextlib
+from threading import Lock
```
tacaswell: Is this general paranoia or are you using threads?

pims/cine.py
```diff
@@ -520,7 +522,12 @@ def __len__(self):
 
     def get_time(self, i):
         '''Returm the time of frame i in seconds.'''
-        return float(i) / self.frame_rate
+        # TODO: This is not guaranteed to be the actual time.
```
tacaswell: Guess we never used the external sync feature in the lab.

I have some recollection of Dustin reporting that if you changed the frame rate mid-collection the images came out garbled (I assume due to electronics not liking their clock rates changed quickly).

pims/norpix_reader.py
```diff
@@ -0,0 +1,269 @@
...
+Length: {count} frames
+Frame Shape: {w}w x {h}h
+Pixel Datatype: {dtype}""".format(filename=self.filename,
+                                  count=len(self),
+                                  w=self.frame_shape[0],
+                                  h=self.frame_shape[1],
+                                  dtype=self.pixel_type)
+
+
+class FileLocker(object):
```
tacaswell: Shouldn't this be shared with the cine code? (or did I take it out when I ported it over to pims?)

pims/base_frames.py
nkeim: I saw that the Cine code was meant to be thread safe, so I just followed that example. I take it that this was never a project-wide goal? If not, perhaps we should replace all the locking stuff with `# NOTE: Not thread-safe`.

pims/cine.py
nkeim: At high frame rates this is probably true. At low frame rates the cameras really didn't seem to care.

pims/norpix_reader.py
nkeim: Yes, it's in cine.py as well. It's such a dead-simple class that I don't think we'd gain much from de-duplication, plus we'd have to create something like `utils/misc.py`. But I'll do it if you think I'm just being lazy :).

pims/base_frames.py
tacaswell: The threading stuff came from Dustin who was doing some stuff with threads.   I would say we leave it in and document it as 'probably thread safe'.

pims/norpix_reader.py
tacaswell: It would be better to not have this in two places, saves the future headache if we want to make this class smarter.

At least add a note that it is duplicated?

If I were doing this I would make a new file to share it, but I don't care enough to put a PR on top of yours to do it.

nkeim:
Implemented @tacaswell 's suggestions and got Travis to pass. (Also rebased to current master.)

danielballan:
Looks good. Since this and #158 are interdependent I leave to you to merge. Merge at will!

nkeim:
Thanks. Due to the level of discussion and refinement that went on here, it's more productive to just merge this branch; I've closed #158 without merging. 

Since @danielballan 's approval, I updated a couple docstrings that I missed earlier and tidied up the history, so I'd be happiest if someone else made this final. Thanks!

danielballan:
One last look, @tacaswell?

tacaswell:
I am not super stoked about the 1MB file.  Can we get LFS set up for this repo?

In the mean time can we expunge that file from the history and change the tests that use it to check if they can find the file and if not raise `SkipTest` (or KnownFail).

nkeim:
@tacaswell , if I could get it down to ~100 KB, would that be better? I'm worried that "not always tested" could lead to "gee, it works for me."

tacaswell:
at 100k I would grumble, but be ok.

I work off of 4G teethering a lot so I am really sensitive to repo size.

nkeim:
Understood. I'll see what I can do. Thanks!

nkeim:
I got it down to 57 KB, and edited history so that the old 1 MB file should never end up in the soft-matter repo. I also added a couple tests based on the specific contents of the file, which seemed like a good idea. I think that wraps it up.

danielballan:
Thanks for humoring the data-thrifty of your train-commuting collaborators.

PR #163, opened 2015-07-13
ENH: Make ImageSequenceND work with RGB files

caspervdw:
Many images are RGB; ImageSequenceND was not working with these images. This addition checks the number of dimensions in the first frame and interprets the RGB dimension as the 'c' dimension. The rest of the functionality is the same as before.

It is not possible to output interleaved RGB images (axes order `yxc`) due to limitations of `FrameSequenceND`. If the user wants to output this, `bundle_axes` should be put to `cyx` and then use `np.rollaxis`. Downside is that in this way, the full RGB image is read 3 times, each time dropping all but one channel.

- API change: `dim_identifiers` is now `axes_identifiers`
- additionally renamed dimension to axes everywhere within `FrameSequenceND` scope
- add tests for ImageSequenceND with RGB (interleaved) PNGs

### DIFF
--- a/pims/image_sequence.py
+++ b/pims/image_sequence.py
@@ -271,10 +271,15 @@ class ImageSequenceND(FramesSequenceND, ImageSequence):
         This determines which axes will be bundled into one Frame. The axes in
         the ndarray that is returned by get_frame have the same order as the
         order in this list. The last two elements have to be ['y', 'x'].
-        Defaults to ['z', 'y', 'x'].
-    default_coords: dict of int
+        If the 'z' axis exists then it defaults to ['z', 'y', 'x']
+    default_coords : dict of int
         When an axis is not present in both iter_axes and bundle_axes, the
         coordinate contained in this dictionary will be used.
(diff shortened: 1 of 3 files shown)


pims/image_sequence.py
```diff
@@ -307,12 +333,22 @@ def get_frame(self, i):
...
+            row = [ind[name] for name in self.axes_identifiers if name != 'c']
+        else:
+            row = [ind[name] for name in self.axes_identifiers]
         i = np.argwhere(np.all(self._toc == row, 1))[0, 0]
         res = self.imread(self._filepaths[i], **self.kwargs)
         if res.dtype != self._dtype:
             res = res.astype(self._dtype)
-        return res
+        if self.isRGB:
+            if self.isInterleaved:
```
danielballan: nitpick: Can we make this the PEP8-compliant `is_interleaved`?

caspervdw:
Done, also added docstrings for the new names `is_rgb` and `is_interleaved`

PR #165, opened 2015-07-20
Do not flip and transpose TIFF images

### LINKED ISSUE #164: TIFF stacks are rotated 90° counterclockwise
At least with the PIL and tifffile backends (I couldn't get libtiff to work), all images appear to be rotated counterclockwise by 90°. If, for instance, I split the TIFF stack into single JPEGs (using e. g. ImageJ) and load them into an ImageSequence, the images are not rotated.

I guess this weirdness has to do with all the flipping and transposing in the get_frame functions.

lschr:
as this results in a counterclockwise rotation by 90°. Apply to
TiffStack_pil and TiffStack_tifffile. Couldn't test TiffStack_libtiff since I
couldn't get it to work.

This should fix #164.

### DIFF
--- a/pims/tests/test_common.py
+++ b/pims/tests/test_common.py
@@ -629,8 +629,8 @@ class TestTiffStack_pil(_tiff_image_series):
 
     def setUp(self):
         self.filename = os.path.join(path, 'stuck.tif')
-        self.frame0 = np.load(os.path.join(path, 'stuck_frame0.npy')).T[::-1]
-        self.frame1 = np.load(os.path.join(path, 'stuck_frame1.npy')).T[::-1]
+        self.frame0 = np.load(os.path.join(path, 'stuck_frame0.npy'))
+        self.frame1 = np.load(os.path.join(path, 'stuck_frame1.npy'))
         self.klass = pims.TiffStack_pil
         self.kwargs = dict()
         self.v = self.klass(self.filename, **self.kwargs)
@@ -644,8 +644,8 @@ class TestTiffStack_tifffile(_tiff_image_series):
 
     def setUp(self):
         self.filename = os.path.join(path, 'stuck.tif')
-        self.frame0 = np.load(os.path.join(path, 'stuck_frame0.npy')).T[::-1]
-        self.frame1 = np.load(os.path.join(path, 'stuck_frame1.npy')).T[::-1]
+        self.frame0 = np.load(os.path.join(path, 'stuck_frame0.npy'))
+        self.frame1 = np.load(os.path.join(path, 'stuck_frame1.npy'))
         self.klass = pims.TiffStack_tifffile
         self.kwargs = dict()
         self.v = self.klass(self.filename, **self.kwargs)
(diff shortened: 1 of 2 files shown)


tacaswell:
This is an API break, people have code that relies on the current orientation

lschr:
I just tested TiffStack_tifffile. It does not rotate the images. So, depending on what Tiff plugin PIMS picks behind the scenes, one gets a different result — very bad.

Also, this rotation is inconsistent with the fact that for other readers such as ImageSequence the images are not rotated.

If people's code relies on the image orientation being wrong, this code is nothing but a workaround around #164, which I would not call a very strong reason not to fix it.

lschr:
EDIT: I tested TiffStack_libtiff, not TiffStack_tifffile, of course.

tacaswell:
That may be the case, but it is still very bad form to break users analysis code.  We have this all the time it mpl, there is code that is _wrong_ but not broken.  From the point of view of users, 'fixing' the bug is introducing a new bug in that their code stops working.  Like doctors, the prime directive is to do no harm.

I am the source of some (all?) of those flips, but no longer remember why they are there.  I am a bit concerned that there maybe setting in tiffs to store the data row major vs column major (or something related to that general bastion of pain).

lschr:
Is this (potential) breaking of user code really an issue? Who uses git master (and blindly pulls from there) in a production environment?

Besides, isn't there a major API overhaul planned anyways (getting rid of process_func and friends in favor of the pipeline decorator)?

Btw., currently not even TiffStack's frame_shape and the individual Frame's shapes agree.

tacaswell:
OK, the mismatched shape is a problem that needs to be fixed.

Pretty sure this is behavior is there is a release, not just master.

lschr:
Right. So people depending on the rotation can simply stick to the release they are using now. Nothing breaks for them. The bug can be fixed in master and eventually be released with a note that the release contains this fix and may need some extra attention. However, no one is forced to upgrade.

tacaswell:
Sorry for the silence on this, I am now convinced this _is_ a bug, not intended behavior but have not had the bandwidth to look into it.

danielballan:
I'm convinced. To be sure, I made two demo tiff stacks, color and grayscale variants of the words "This Side Up." I read it with the three tiff readers using current master and this PR.

* [Current Behavior](http://nbviewer.ipython.org/gist/danielballan/a35bee4fa06dcb09b3b0)
* [New Behavior Proposed in this PR](http://nbviewer.ipython.org/gist/danielballan/32095e6f329efc002d26)

With @lschr's changes, all readers show the image in the correct orientation. They should never have been rotated in the first place. I think I put that in while wrestling with a weird format in my early days of learning Python. Later, @tacaswell dutifully propagated my code, and here we are. Sorry that it caused headaches.

This change also makes `v[0].shape` and `v.frame_shape` internally consistent (which we should cover with unit tests!) and shows the image in the correct orientation.

As a side note, these notebooks show that our PIL reader can't read color images. @tacaswell or I will fix that in a separate PR.

This may break user code, but it's the right thing thing to do.

danielballan:
P.S. @lschr This also needs a rebase.

lschr:
Hmmm, the python 2.7 tests fail, but I don't think that has to do with my commits.

danielballan:
It was just a conda hiccup. I re-ran the build and all's well. Merging!

PR #166, opened 2015-07-22
Add reader vor SPE files

lschr:
Apart from the actual image data, also some metadata from the file header is
read and stored in a dict named "metadata". Not all metadata fields are
currently read from a file (but reading of additional fields is very easy to
add). Reading XML metadata as per SPE 3.0 specification is not supported.


tacaswell:
How does this compare with the reader at https://github.com/tacaswell/filestore/blob/master/filestore/readers/spe.py ?

tacaswell:
and are these even the same type of spe file?

tacaswell:
Sorry, I should have led with "Thank you!"

lschr:
> How does this compare with the reader at https://github.com/tacaswell/filestore/blob/master/filestore/readers/spe.py ?
A quick look suggests that these are the same SPE files.
The main differences seem to be (if I am not mistaken, I am not familiar with the code you pointed me to)
- The implementation from filestore loads the whole file into RAM
- The filestore code saves metadata as attributes, while mine uses a dict

Also there are some minor differences like what metadata is read, support for process_func and dtype etc.

tacaswell:
From that description I like your version better!  How much work would it be make this version have the union of the metadata extracted?

Related to xml, we have some (currently not working?) example of how to do this at https://github.com/soft-matter/pims/blob/master/pims/tiff_stack.py#L372

lschr:
Adding more metadata is very simple. All it takes is adding (offset, dtype, count) tuples to the Spec.metadata dict. The reason why I did not do that (yet) is testing: Our software only writes a subset of the metadata, and I only implemented those (and the SPE specs are not very helpful in figuring stuff out unless one can try oneself). I guess, adding the additional entries from the filestore reader would take about ten minutes.

However, there is the question of naming: I stuck with the names from the specs. The filestore version uses more descriptive names.

Last but not least: The XML metadata in the SPE 3.0 spec describes the layout of binary data in the file, and not only some additional information. This would require a completely new reader class to implement.

tacaswell:
Ah, sorry for the xml confusion.

I think stick with the names in the spec.

lschr:
OK, done.

danielballan:
@lschr This looks good. Would you rebase the branch onto master so I can merge?

lschr:
Here you go :-)

danielballan:
Thanks again. I'm taking a closer look at this TIFF issue now, will update later today.

PR #168, opened 2015-07-25
TiffStack_*: Read (some) metadata

lschr:
The following are read: ImageDescription, DateTime, Software, and
DocumentName.


danielballan:
Awesome. Thank you for doing this. Do you have time to add some tests to cover this behavior?

lschr:
Done.

danielballan:
Thanks again!

Issue #173, opened 2015-08-21
High aspect ratio Frame repr


caspervdw:
As a note to self:

Sometimes it happens that you have a picture of high aspect ratio (e.g. (512, 4)). The rich representation will stretch this so that the width is 512 (and height 65536). This may crash your browser and definitely doesn't look well. There should be some safety value for the height.

caspervdw:
Closing due to PR #174

PR #174, opened 2015-08-26
FIX Height and depth safety checks for frame repr

caspervdw:
This solves issue #173

For rich IPython display of frames, the image was always resized so that width would be 512. Now there is an additional check that ensures height to stay below 512.

Also, for scrollable stacks, the max depth is set to 128 because of long waiting times of generating a large scrollable stack.

### DIFF
--- a/pims/frame.py
+++ b/pims/frame.py
@@ -9,6 +9,8 @@ from pims.display import _scrollable_stack, _as_png, to_rgb
 
 
 WIDTH = 512  # width of rich display, in pixels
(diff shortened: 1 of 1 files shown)


danielballan:
Poke Travis again?

caspervdw:
Done, I recently pushed a second commit that completely rewrites the logic. The output didn't really change but it is much clearer to understand. I couldn't wrap my had around the current code and there were still issues with images of (for instance) shape (2, 512, 512, 3).

danielballan:
Agreed that that is clearer.

PR #177, opened 2015-08-27
FIX Bugfix zero out default coords of bundled axes

caspervdw:
Another bugfix.. I hope this can be merged before tagging 0.3.

FramesSequenceND gave unpredictable results when the `default_coord` of an axis that is bundled was nonzero. To get rid of this, the bundled axes are zeroed before the frame is composed in `get_frame`.


danielballan:
Now that the tifffile fix is merged, would you poke Travis to re-run?

caspervdw:
Fixed

PR #179, opened 2015-09-08
FIX Plots to frame pad mode

caspervdw:
This fixes a technical issue in python 3. `np.pad` takes a mode that is a string, which should be unicode in py3 and bytes in py2.7. Currently (using `b'constant'`) it failed in python 3, and together with `from __future__ import unicode_literals` it does need to be converted in Py2.7. The solution is a `str()` operation.

This fixes also `tp.annotate3d` under python 3.


caspervdw:
Error is related to the Norpix reader, see #180

PR #183, opened 2015-09-18
Rebase on slicerator

caspervdw:
This rebases the pims SliceableIterable on slicerator.

Test do not pass because of 2 issues:
- pypi doesn't have the latest version of slicerator yet (waiting for https://github.com/soft-matter/slicerator/pull/4)
- the pipelines do not work

The pipeline does raise an interesting issue: it only functions on `Slicerator`s while it should also function on the base `FramesSequence` object. So I should write a custom pims pipeline that also works on `FramesSequence` objects.

@danielballan wouldn't it make more sense to construct the Slicerator by subclassing it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - conda install -n testenv -c soft-matter --yes pyav tifffile
   - source activate testenv
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then pip install libtiff; fi
-  - pip install jpype1;
+  - pip install slicerator jpype1;
   - python setup.py build_ext install
 
 
--- a/pims/api.py
+++ b/pims/api.py
@@ -1,8 +1,8 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-from pims.base_frames import (FramesSequence, FramesSequenceND,
-                              SliceableIterable, pipeline)
+from slicerator import pipeline
+from pims.base_frames import FramesSequence, FramesSequenceND
 from pims.frame import Frame
 from pims.display import (export, play, scrollable_stack, to_rgb, normalize,
                           plot_to_frame, plots_to_frame)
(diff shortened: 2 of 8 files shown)


danielballan:
> wouldn't it make more sense to construct the Slicerator by subclassing it?

This is tricky. I would it to be possible to use Slicerator without defining a class; that seems too heavy for general utility. However, I agree that that is natural way to construct a Slicerator.

It would be ideal if `FramesSequence` and all the PIMS readers returned an object that *is* a Slicerator (as opposed to *has* a Slicerator). Could we make `FramesSequence` inherit from `Slicerator`? Multiple inheritance does not play well with abstract base classes, so that might be tricky and/or involve discarding `FramesStream`.

We could revise the Slicerator PR so that `Slicerator` must be subclassed to define a `get` method. Then we could provide factory functions that conveniently define a subclass based on a function/list/etc.

The number of ways to approach this is paralyzing!

danielballan:
I re-ran the tests, but Travis is failing on a deep error that looks like a scipy installation issue. Running the tests locally, I get errors of this kind from all the bioformats tests.

```
======================================================================
ERROR: test_metadata_raw (pims.tests.test_bioformats.TestBioformatsMetadataND2)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/dallan/Documents/Repos/pims/pims/tests/test_bioformats.py", line 529, in test_metadata_raw
    self.v = self.klass(self.filename, meta=False)
  File "/Users/dallan/Documents/Repos/pims/pims/bioformats.py", line 334, in __init__
    log4j.BasicConfigurator.configure()
  File "/Users/dallan/miniconda/envs/py3/lib/python3.4/site-packages/jpype/_jpackage.py", line 55, in __call__
    raise TypeError("Package {0} is not Callable".format(self.__name))
TypeError: Package org.apache.log4j.BasicConfigurator.configure is not Callable
```

caspervdw:
This bioformats issue, you might miss log4j or have the wrong version. Without going into the java details, We could catch the exception and dont use the logger if it is missing.

PR #184, opened 2015-10-11
Slicerator rebase -- debugging Travis

danielballan:
This is #183 plus some commits to debug Travis, probably related to the recent release of numpy 1.10.


danielballan:
This is Casper's PR plus some commits from me that only touch .travis.yml. Hurray for slicerator! Thanks for doing the rebase, @caspervdw.

PR #185, opened 2015-10-12
Fix deps

danielballan:
This setup.py is updated to include the dependencies that we *intend* to be required. Some others are accidentally required -- this will be fixed in a separate PR.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include MANIFEST.in
 include versioneer.py
-include pims/_version.py
+include pims/_slicerator_version.py
 include setup.py
 include setup.cfg
 include license.txt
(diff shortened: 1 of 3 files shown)


caspervdw:
@danielballan are you OK with merging the slicerator dependency now?

The 0.3 tag does not list this dependency so that every user will see an `ImportError` until he/she installs slicerator.

danielballan:
Fair enough. I wanted to fix the long-existing problem with scipy and PIL deps too, but I'll merge this and push a v0.3.1. Tags are cheap. :- )

danielballan:
Done! Sorry for the runaround, @caspervdw.

danielballan:
Sigh. Caught you correct comment too late. I tagged and upload v0.3.2rc1. If it works I'll make it v0.3.2.

caspervdw:
Sorry for my late comment and good idea to make it an rc!

Now trackpy build fails because we dropped the PIL dependency of trackpy. Seems like #186 needs to be addressed soon.

danielballan:
Agreed.

PR #186, opened 2015-10-12
WIP: Split ImageSequence into a reader per backend.

danielballan:
This is still a work in progress. Just pushing so I don't lose or forget about this old work.

### DIFF
--- a/pims/api.py
+++ b/pims/api.py
@@ -13,7 +13,7 @@ import os
 from warnings import warn
 
 # has to be here for API stuff
-from pims.image_sequence import ImageSequence, ImageSequenceND  # noqa
+from pims.image_sequence import ImageSequenceND  # noqa
 from .cine import Cine  # noqa
 from .norpix_reader import NorpixSeq  # noqa
 from pims.tiff_stack import TiffStack_tifffile  # noqa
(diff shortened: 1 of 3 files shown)


caspervdw:
What's your opinion on splitting the two changes:
- make PIL/scipy optional
- reduce the number of kwargs

Then we can first fix the deps, and then break the API in 0.4

We could do all the import-checks inside `utils` and make use of them throughout pims (for instance in `display.py` there is also a PIL and mpl dependency)

pims/image_sequence.py
```diff
@@ -225,7 +265,7 @@ def filename_to_indices(filename, identifiers='tzc'):
     return result
 
 
-class ImageSequenceND(FramesSequenceND, ImageSequence):
+class ImageSequenceND(FramesSequenceND, BaseImageSequence):
```
caspervdw: I think it makes sense to make this `BaseImageSequenceND`

pims/image_sequence.py
```diff
@@ -349,8 +389,8 @@ def customize_image_sequence(imread_func, name=None):
     >>> MyImageSequence = customize_image_sequence(my_func)
     >>> frames = MyImageSequence('path/to/my_weird_files*')
     """
-    class CustomImageSequence(ImageSequence):
-        def imread(self, filename, **kwargs):
+    class CustomImageSequence(BaseImageSequence):
+        def _imread(self, filename, **kwargs):
             return imread_func(filename, **kwargs)
     if name is not None:
         CustomImageSequence.__name__ = name
```
caspervdw: `__doc__ = BaseImageSequence.__doc__`

pims/image_sequence.py
```diff
@@ -17,19 +17,52 @@
...
+    from PIL import Image
+except ImportError:
+    pil_imread = None
+else:
+    def pil_imread(filename):
+        return np.asarray(Image.open(filename))
+
+
+def skimage_available():
+    return skimage_imread is not None
```
caspervdw: If you move this to `utils` then we can make use of this  in `display.py` too

pims/api.py
```diff
@@ -57,6 +57,33 @@ def raiser(*args, **kwargs):
     TiffStack = not_available("tifffile, libtiff, or PIL/Pillow")
 
 
+# As above
+from pims.image_sequence import (ImageSequence_skimage, ImageSequence_mpl,
+                                 ImageSequence_scipy, ImageSequence_pil,
+                                 ImageSequence_tifffile)
+if not pims.image_sequence.skimage_available():
+    ImageSequence_skimage = not_available('skimage')
+if not pims.image_sequence.skimage_available():
```
caspervdw: I would opt for an (optional) classmethod `pkg_available()` for every reader class.

pims/image_sequence.py
```diff
@@ -73,19 +102,8 @@ class ImageSequence(FramesSequence):
     >>> frame_shape = video.frame_shape # Pixel dimensions of video
     """
     def __init__(self, path_spec, process_func=None, dtype=None,
```
caspervdw: If this works:

    @classmethod
    def pkg_available(cls):
        return cls._imread is not None

PR #188, opened 2015-10-23
Drop PIL and Scipy dependencies

caspervdw:
This makes PIL and scipy really optional, as a step up to #186 . I added a Travis mode with `numpy` only.

The PyAv reader is based on PIL too, so there is an extra check for that now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,18 @@ sudo: false
 matrix:
   include:
     - python: "2.7"
-      env: DEPS="numpy=1.9*"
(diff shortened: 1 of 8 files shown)


danielballan:
I read it twice, looks good. We'll settle the pillow v3 issue in the next round.

Do you want an v0.3.2rc2?

caspervdw:
Yes please then trackpy tests can pass again

danielballan:
Done and pushed to PyPI. Let me know if this looks good and I'll tag v0.3.2
final.

Dan

PR #190, opened 2015-10-26
DOC Pypi readme conversion

caspervdw:
@danielballan  I just found out this code snippet, it configures the readme.md to rst which is displayed nicely on pypi. Just `pip install pypandoc` and it should work.


danielballan:
Slick.

Issue #191, opened 2015-11-02
Dropping FramesSequenceMappable


caspervdw:
Before releasing 0.3.2, shall we drop FramesSequenceMappable? It should be superseded by slicerator.

caspervdw:
@nkeim I should address you for this.

nkeim:
Yes. See my PR.

PR #192, opened 2015-11-02
CLN: Drop FramesSequenceMappable (replaced by slicerator)

nkeim:
Addresses #191

### DIFF
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -6,15 +6,10 @@ from six import with_metaclass
 from six.moves import range
 import os
 import numpy as np
-import collections
 import itertools
-import functools
-import contextlib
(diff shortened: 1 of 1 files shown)


pims/base_frames.py
```diff
@@ -11,7 +11,6 @@
 import functools
 import contextlib
```
caspervdw: This is also superfluous

caspervdw:
Thanks. Couls you also drop the _index_generator? It appears that I missed it before.

nkeim:
I removed all of the unused module imports (except `six`), and `_index_generator`.

danielballan:
Thanks for the cleanup! `_index_generator` still gives me nightmares.

PR #194, opened 2015-11-05
FIX Image import in plot_to_frame

caspervdw:
A missing import in plot_to_frame, introduced in #188


danielballan:
Good catch. We should improve test coverage of display.py.

PR #195, opened 2015-11-05
FIX: In repr, do not assume frame_shape is 2D.

danielballan:
I'd also like this for v0.3.2.

@caspervdw please review/merge and let me know if we are otherwise ready for a tag.


caspervdw:
:+1: looks good, ready for tag 0.3.2

danielballan:
Here we gooooooo

PR #199, opened 2015-11-12
FIX/TST: Fix zipfile handling on py3; add test.

### LINKED ISSUE #198: BUG: zipfile handling does not work in py3
I discovered a bug while updating the tracking-3d tutorial. As far as I can tell there is no test coverage of the zipfile feature, so I can believe that this slipped by us.

I tried the obvious fix -- adding `str` -- and that did not work. Will keep thinking about this.

```python
In [11]: pims.ImageSequenceND(r'/Users/dallan/Documents/Repos/trackpy-examples/sample_data/pmma_colloids.zip', axes_identifiers = ['z', 't'])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-689e411c9c33> in <module>()
----> 1 frames = pims.ImageSequenceND(r'/Users/dallan/Documents/Repos/trackpy-examples/sample_data/pmma_colloids.zip', axes_identifiers = ['z', 't'])

/Users/dallan/Documents/Repos/pims/pims/image_sequence.py in __init__(self, path_spec, process_func, dtype, as_grey, plugin, axes_identifiers)
    297         self.axes_identifiers = axes_identifiers
    298         super(ImageSequenceND, self).__init__(path_spec, process_func,
--> 299                                               dtype, as_grey, plugin)
    300         shape = self._first_frame_shape
    301         if len(shape) == 2:

/Users/dallan/Documents/Repos/pims/pims/image_sequence.py in __init__(self, path_spec, process_func, dtype, as_grey, plugin)
     93         self._get_files(path_spec)
     94 
---> 95         tmp = self.imread(self._filepaths[0], **self.kwargs)
     96         self._first_frame_shape = tmp.shape
     97 

/Users/dallan/Documents/Repos/pims/pims/image_sequence.py in imread(self, filename, **kwargs)
    118                               "scipy, matplotlib or scikit-image.")
    119         if self._is_zipfile:
--> 120             file_handle = StringIO(self._zipfile.read(filename))
    121             return imread(file_handle, **kwargs)
    122         else:

TypeError: initial_value must be str or None, not bytes
```

danielballan:
Closes #198

### DIFF
--- a/pims/tests/test_common.py
+++ b/pims/tests/test_common.py
@@ -4,12 +4,14 @@ from __future__ import (absolute_import, division, print_function,
 import six
 
 import os
+import tempfile
 import zipfile
 import sys
 import random
 import types
 import unittest
 import pickle
+from io import BytesIO
 import nose
 import numpy as np
 from numpy.testing import (assert_equal, assert_allclose)
@@ -571,7 +573,10 @@ class TestImageSequenceWithPIL(_image_series, unittest.TestCase):
         self.v = self.klass(self.filename, **self.kwargs)
         self.expected_shape = shape
         self.expected_len = 5
-        with zipfile.ZipFile('test.zip', 'w') as archive:
+        self.tempdir = tempfile.mkdtemp()
+        self.tempfile = os.path.join(self.tempdir, 'test.zip')
+
+        with zipfile.ZipFile(self.tempfile, 'w') as archive:
             for fn in self.filenames:
                 archive.write(os.path.join(self.filepath, fn))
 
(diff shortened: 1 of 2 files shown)


pims/tests/test_common.py
```diff
@@ -570,13 +571,20 @@ def setUp(self):
         self.v = self.klass(self.filename, **self.kwargs)
         self.expected_shape = shape
         self.expected_len = 5
+        with zipfile.ZipFile('test.zip', 'w') as archive:
```
tacaswell: Can you use a tempdir/tempfile https://docs.python.org/3.5/library/tempfile.html for this?

danielballan: Add a commit or paint a bikeshed. ;- ) Don't trust setup/teardown in nose?

tacaswell: No, I don't trust that this won't run in a directory where there is an existing user file called 'test.zip'

danielballan: Good point. Fixed.

PR #203, opened 2015-11-19
Moviepy and ImageIO

caspervdw:
This contains some older work that I am already using for some time now:

- video readers via ImageIO and MoviePy , both packages have different implementations of an ffmpeg-based video reader via pipes
- video exporter via moviepy, similar api as the PyAv exporter
- tests for the readers

One thing, the PyAV reader seems to be tested with an expected shape that is (x, y) in stead of the conventional (y, x). @danielballan  was this intentional?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,21 +4,20 @@ sudo: false
 matrix:
   include:
     - python: "2.7"
-      env: DEPS="numpy=1.9*" DEPSSM="tifffile"
+      env: DEPS="numpy=1.9*" DEPSSM="tifffile" DEPSPIP="slicerator libtiff"
     - python: "2.7"
-      env: DEPS="numpy=1.9* scipy pillow matplotlib scikit-image jinja2" DEPSSM="pyav tifffile"
+      env: DEPS="numpy=1.9* scipy pillow matplotlib scikit-image jinja2" DEPSSM="pyav tifffile" DEPSPIP="slicerator libtiff moviepy imageio jpype1"
     - python: "3.4"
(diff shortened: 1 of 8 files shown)


danielballan:
This work probably supercedes my old work in https://github.com/soft-matter/pims/blob/master/pims/ffmpeg_reader.py

caspervdw:
There is a thing I already discussed shortly with @danielballan on Skype. ImageIO contains a very nice  infrastructure for keeping track of all readers (`Format` classes that contain `Reader` and `Writer` classes). Each `Format` is registered in the `FormatManager` that selects the right reader for each file. `Format` exposes a `can_read` method that (optionally) includes the interpretation of the first 256 bytes of the file.

https://github.com/imageio/imageio/blob/master/imageio/core/format.py

Because this infrastructure exists already, I am considering the option of using it inside PIMS and making all of our readers ImageIO-compatible (which is not too bad because APIs are much alike).

This will include either copying over code or making imageio a required dependency.

What do you think about this?

danielballan:
If we made imageio a required dep, what secondary required deps would that pull in?

caspervdw:
None, the only dependency of imageio is Numpy. It downloads binaries of for instance Freeimage from their git repo when required.

caspervdw:
Anyone :-1: on accepting ImageIO as a required dependency?

tacaswell:
go for it.

danielballan:
:+1:

danielballan:
This needs a rebase, probably just in .travis.yml.

caspervdw:
I'll close this because we chose for a different approach that makes imageio a required dep and makes use of the imageio reader selection methods.

PR #204, opened 2015-11-24
FIX Pillow 3.0.0 compatibility

caspervdw:
This circumvents PIL, now there should not be an issue with PIL 3.0.0.

Before, the code used `fig.imsave` to a buffer, now it uses the built in `print_to_buffer` from the `FigureCanvasAgg` backend. I still need to look at how to correctly pass the kwargs that I used to pass into imsave. For instance `bbox_inches='tight'` that trackpy uses in `annotate3d`.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -440,20 +440,19 @@ def plots_to_frame(figures, width=512, close_fig=False, bbox_inches=None):
         raise ValueError('Use plot_to_frame for single figures, or supply '
                          'an iterable of figures to plots_to_frame.')
 
-    # render first image to calculate the correct dpi and image size
-    size = plot_to_frame(figures[0], 100).shape
-    dpi = width * 100 / size[1]
-    h = int(width * size[0] / size[1])
     width = int(width)
+    h = None
 
     frames = []
     for n, fig in enumerate(figures):
-        im = plot_to_frame(fig, dpi, close_fig)
-        # make the image the same size as the first image
-        if (im.shape[0] != h) or (im.shape[1] != width):
-            im = np.pad(im[:h, :width], ((0, max(0, h - im.shape[0])),
-                                         (0, max(0, width - im.shape[1])),
-                                         (0, 0)), mode=str('constant'))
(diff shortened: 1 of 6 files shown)


danielballan:
There is still *some* issue with PIL 3.0.0, in `TiffStack_pil`. That's the test that fails. You might already be clear on that point -- I'm just jotting this down for the record on GitHub.

I leave it to @tacaswell to review this intricate dance with matplotlib internals. I suspect that the use of `fig.canvas.switch_backends` might be problematic.

pims/display.py
```diff
@@ -392,22 +393,31 @@ def plot_to_frame(fig, dpi, **imsave_kwargs):
...
-    fig.savefig(buffer, format='tif', dpi=dpi, **imsave_kwargs)
-    buffer.seek(0)
-    im = np.asarray(Image.open(buffer))
-    buffer.close()
-    return Frame(im)
+
+    agg = fig.canvas.switch_backends(mpl.backends.backend_agg.FigureCanvasAgg)
+    original_bbox = fig.bbox_inches
+    if bbox_inches is not None:
+        fig.bbox_inches = bbox_inches
```
caspervdw: This doesn't seem to have an effect. @tacaswell do you no a way to use a tight bounding box in this method? `savefig(..., bbox_inches='tight')` gave nicer results than this implementation.

caspervdw:
@danielballan  fixed the compatibility issue in `TiffStack_pil`.

danielballan:
Nice work! The tiff fix looks good. I appreciate the tests also.

caspervdw:
Mentioning https://github.com/caspervdw/pims/pull/1

danielballan:
Note: I merged this too early, realized Tom's commit was missing, merged his commit locally and pushed to master. All is well now.

caspervdw:
Ok thanks for doing this, I almost made a new PR. 

I think that it still needs some improvement as `bbox_inches='tight'` that `annotate3d` uses will now raise an error. Also I am not sure if `bbox_inches` should actually *resize* the figure, I thought that it should *crop* the figure.

PR #206, opened 2015-11-25
FIX Plot to frame and annotate3d compatibility

caspervdw:
This builds on on @tacaswell 's adaptions.

The issue that persisted is the following: trackpy gives `bbox_inches = 'tight'` to `plots_to_frame`, this is now failing. We don't want to coordinate a release of trackpy now so we have to solve it.

I did the following changes:

When `bbox_inches = 'tight'`, the layout is set to tight from inside the figure context manager using `fig.set_tight_layout(True)`. The figure resize is now done via `fig_size_inches` kwarg.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -417,8 +417,8 @@ def plot_to_frame(fig, width=512, close_fig=False, fig_size_inches=None,
         The width of the resulting frame, in pixels
     close_fig : boolean
         When True, the figure will be closed after plotting
-    fig_size_inches : tuple of numbers
-        The figure size in inches
+    fig_size_inches : tuple
+        The figure (height, width) in inches. If None, the size is not changed.
     bbox_inches : {'tight', None}
         When 'tight', tight layout is used.
 
@@ -434,7 +434,6 @@ def plot_to_frame(fig, width=512, close_fig=False, fig_size_inches=None,
     if fig_size_inches is not None:
         if fig_size_inches[0] == 0 or fig_size_inches[1] == 0:
             raise ValueError('Figure size cannot be zero.')
-        fig_size_inches = fig_size_inches[:2]
     if bbox_inches is None:
         tight_layout = None
     elif str(bbox_inches) == 'tight':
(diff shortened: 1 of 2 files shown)


pims/display.py
```diff
@@ -420,14 +431,19 @@ def plot_to_frame(fig, width=512, close_fig=False, bbox_inches=None):
     from pims import Frame
     if isinstance(fig, mpl.axes.Axes):
         fig = fig.figure
+    if fig_size_inches is not None:
```
tacaswell: If you are going to rename this, might as well make it a len 2 tuple

caspervdw: It was already, I put it in the docs. It can be a len 2 tuple or None, in which case the figsize is unchanged.

caspervdw:
Is this PR now ok for you @tacaswell  ?

pims/display.py
```diff
@@ -387,20 +387,26 @@ def _fig_size_cntx(fig, fig_size_inches):
...
+        fig.set_tight_layout(tight_layout)
     try:
         yield fig
     finally:
         fig.set_size_inches(*orig_size)
+        fig.set_tight_layout(orig_layout)
 
 
-def plot_to_frame(fig, width=512, close_fig=False, bbox_inches=None):
+def plot_to_frame(fig, width=512, close_fig=False, fig_size_inches=None,
```
tacaswell: This is an API break if anyone is using positional arguments (I think my source of confusion below was seeing that the 3rd positional arg had been renamed).  I am :+1: on making this change, but it should be documented.

tacaswell:
LGTM modulo the API break documentation.

pims/display.py
caspervdw: There was already an API break in #204; in the latest release (v0.3.2) it has the following arguments: `fig, dpi, **imsave_kwargs`. I will document that.

caspervdw:
We do not have a logistic to put release notes somewhere. What is the best way to do that?

@tacaswell OK the merge this and solve the release notes in one go?

PR #207, opened 2015-11-26
ENH pims.open checks FramesSequenceND

caspervdw:
There are a couple of changes in `pims.open` here, partly already aiming at removing the kwargs from the readers:

- subclasses of `FramesSequenceND` are also listed
- `pims.open` takes `kwargs` and forwards them to the reader, without any inspection
- additionally, it iterates over all handlers and returns when one did not error
- tests for `open` are done for all bioformats formats. Sometimes it indeed tries the `TiffStack` first and then `Bioformats`.
- bioformats listed the extentions as `'.tif'` instead of `'tif'`, fixed that.

### DIFF
--- a/pims/api.py
+++ b/pims/api.py
@@ -127,7 +127,7 @@ def open(sequence, **kwargs):
                          _recursive_subclasses(FramesSequenceND))
     # keep handlers that support the file ext. use set to avoid duplicates.
     eligible_handlers = set(h for h in all_handlers
-                            if ext and ext in _drop_dot_iter(h.class_exts()))
+                            if ext and ext in map(_drop_dot, h.class_exts()))
     if len(eligible_handlers) < 1:
         raise UnknownFormatError(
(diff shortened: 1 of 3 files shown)


pims/api.py
```diff
@@ -172,3 +164,12 @@ def _recursive_subclasses(cls):
     # Source: http://stackoverflow.com/a/3862957/1221924
     return (cls.__subclasses__() +
         [g for s in cls.__subclasses__() for g in _recursive_subclasses(s)])
+
+def _drop_dot(s):
+    if len(s) == 0 or s[0] != '.':
```
danielballan: There is also `s.startswith('.')` for readability and probably a minor performance advantage.

danielballan:
Looks good. See minor line note (which you are free to ignore...) and merge at will.

caspervdw:
@danielballan I made the code neater using `startswith` and `map`. You can merge after Travis passes.

Issue #215, opened 2015-12-14
Feature request: Norpix color images


dmopalmer:
Currently, Norpix .seq files don't support color camera files.  Multi-color camera files raise:
```
Only uncompressed mono images are supported in .seq files
```

nkeim:
Thank you! That must be a frustrating error to receive. As I recall, the `seq` format does not actually store color data, but instead simply encodes the color filter pattern. pims will have to use a color reconstruction algorithm to properly read these files. OpenCV appears to include the appropriate algorithm(s) for the bayer pattern.

This is a pretty low priority for me, but someone else (or you) may wish to take the initiative. One thing that would help is if you could attach here a very small (just 1 or 2 frames, low resolution) color sample file.

caspervdw:
Closing because it is fixed in #218

PR #216, opened 2015-12-14
Support for reading non-monochrome images in NorpixSeq as raw

dmopalmer:
Use the NorpixSeq(filename, as_raw=True) flag.

The user must interpret the image correctly, based on the image_format header
entry to determine how to convert pixel values to RGB, etc.

Compressed files are still not supported.

This is for issue #215

### DIFF
--- a/pims/norpix_reader.py
+++ b/pims/norpix_reader.py
@@ -20,9 +20,10 @@ from threading import Lock
 
 __all__ = ['NorpixSeq',]
 
-DWORD = 'L'
+DWORD = 'L' # unsigned long
 LONG = 'l'
 DOUBLE = 'd'
+USHORT = 'H'
 
 HEADER_FIELDS = [
     ('magic', DWORD),
@@ -41,6 +42,17 @@ HEADER_FIELDS = [
     ('true_image_size', DWORD),
     ('suggested_frame_rate', DOUBLE),
     ('description_format', LONG),
+    ('reference_frame', DWORD),
+    ('fixed_size', DWORD),
+    ('flags', DWORD),
+    ('bayer_pattern', LONG),
+    ('time_offset_us', LONG),
+    ('extended_header_size', LONG),
+    ('compression_format', DWORD),
+    ('reference_time_s', LONG),
+    ('reference_time_ms', USHORT),
+    ('reference_time_us', USHORT)
+    # More header values not implemented
 ]
 
 
@@ -50,7 +62,8 @@ class NorpixSeq(FramesSequence):
     This is the native format of StreamPix software, owned by NorPix Inc.
     The format is described in the StreamPix documentation.
 
-    Currently supports uncompressed 8-bit monochrome files only.
+    Currently supports uncompressed files only, either
+    uint8/16/32 monochrome or as_raw color or monochrome.
 
     Nominally thread-safe.
 
@@ -65,6 +78,11 @@ class NorpixSeq(FramesSequence):
         Image arrays will be converted to this datatype.
     as_grey : boolean, optional
         Ignored.
+    as_raw : boolean, optional
+        Required for non-monochrome frames.
+        Images will be returned as an ndarray of bytes.
+        2-dimensional if the image height evenly divides the bytes per image,
+        1-dimensional otherwise.
     """
     @classmethod
     def class_exts(cls):
@@ -74,7 +92,7 @@ class NorpixSeq(FramesSequence):
                        'get_time_float', 'filename', 'width', 'height',
                        'frame_rate']
 
-    def __init__(self, filename, process_func=None, dtype=None, as_grey=False):
+    def __init__(self, filename, process_func=None, dtype=None, as_grey=False, as_raw = False):
         super(NorpixSeq, self).__init__()
         self._file = open(filename, 'rb')
         self._filename = filename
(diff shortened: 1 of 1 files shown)


dmopalmer:
Do not pull yet.  I will be adding support for non-8-bit in a few hours or days.

nkeim:
Thank you! I agree that this approach is acceptable for now. And higher bit depths would be very welcome.

It would be great to include a unit test for the higher bit depths especially, as that feature will involve multiple parts of the code. Could you generate a very small sample file? If not, I could probably do that.

dmopalmer:
This should now be ready to pull.

(If I did squashing correctly, it should show up as a single commit.)

The as_raw flag now reads any uncompressed image as an array of uint8.

If the number of rows in the image does not evenly divide the number of bytes,
the array will be 1 dimensional, otherwise it will be the height of the image and
the appropriate width in bytes, which is not necessarily the horizontal pixel count.

The user can open the .seq file as_raw, then examine the header and
use set_process_func() to install the correct uint8->image converter.

There are so many possible image formats, and they are so poorly
documented, that I won't try to include any process_func examples
in the repository.

tacaswell:
still see the merge from master which is less than great.

`git rebase -i upstream/master` is your friend :wink:

tacaswell:
Also :+1: for improvements to readers!

dmopalmer:
git rebase -i upstream/master
is my enemy.

Tried to rebase.  Got merge conflicts.  After an hour and a lot of stack overflow etc. I went drastic.
Blew away my fork, reforked, remade my changes and submitted a new pull request: #217.  
That had a typo, and now I'm back to two commits.

https://xkcd.com/1597/

```
% git clone https://github.com/dmopalmer/pims.git
% cd pims
% git remote add upstream https://github.com/soft-matter/pims.git
% git fetch upstream
% git rebase -i upstream/master
[detached HEAD 364782a] Support for non-monochrome and packed-bits images in NorpixSeq, but only with the as_raw flag to return a uint8 array. The user must interpret the raw image correctly, based on the image_format header entry to determine how to convert pixel values to RGB, etc.
 Date: Wed Dec 16 23:39:39 2015 -0700
 1 file changed, 46 insertions(+), 8 deletions(-)
Successfully rebased and updated refs/heads/master.
% git status
On branch master
Your branch and 'origin/master' have diverged,
and have 1 and 2 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
nothing to commit, working directory clean
```

Now what?  
```
% git pull
Merge made by the 'recursive' strategy.
% git status
On branch master
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)
nothing to commit, working directory clean
% git commit
On branch master
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)
nothing to commit, working directory clean
```

Any suggestions?

tacaswell:
Ah, I just iced that this PR is from your master branch, git works better if you work on a feature branch.  One of the problems with git is that it is too powerful/flexible and most of the people answering on SO have found _a_ way to do something but it typically is not the right way, but gets upvoted as a pragmatic (wrong) solution.

```
git remote rename origin dmopalmer
git remote rename upstream soft-matter
```
These two commands rename your remotes because I find 'origin' and 'upstream' super confusing, this is the pattern I use on all of my repos.  One bug/feature of this is that it makes you be explicit about what remote you want to work with as git treats 'origin' as special/automagical.

```
git checkout -b enh_color_norpix
```
creates a new feature branch.  `git branch` should now show that you have two branches `master` and `enh_color_norpix` and you are on `enh_color_norpix`.

```
git rebase -i upstream/master
```
This will rebase your feature branch on top of current upstream

```
git push dmopalmer enh_color_norpix
```
will push a branch without the merges to you gh repo which you should use to open a PR with

To clean up your master (**WARNING POSSIBLY DESTRUCTIVE**)

```
git checkout master
git reset --hard soft-matter/master
git push dmopalmer --force master
```

which will switch you back to the master branch, forcible reset it to match the soft-matter master branch (that is set the commit that you local branch is pointing to to be the same commit as soft-matter/master is pointing to and discard all local changes in your work tree, hence destructive) and then force push to your public repo to do the same there (hence, destructive).

The thing that caused you problems was the `git pull` which was merging back into your local branch the changes you just worked to remove.  I personally do not use `git pull` as it has done things like this to me a few too many times.  My typical workflow is `git remote update` and then `git merge remote/branch --ff-only`.  If you really want to use pull, I think `git pull --ff-only` does what I expect.

git should ship with a program `gitk` which lets you look at the commit tree, that is an invaluable tool.

I also strongly suggest finding a better front end to git than the command line.  I use `emacs` + `magit` which lets me do most of these things from inside of emacs (which is awesome), but if you do not already use emacs, that can be a steep learning curve.  @danielballan assures me there is an equivalent for `vi` and I have heard good things about the GH gui (but it mostly confused me the one time I tried to use it).

hour read, but helped me understand git: http://tom.preston-werner.com/2009/05/19/the-git-parable.html

very quick read explaining the above workflow with pretty pictures: https://guides.github.com/introduction/flow/

dmopalmer:
Thank you.  Try 3 #218 is a single-commit pull request which should be identical to what I want.

But it fails travis when the testing tries to read the test tiff tile from 
https://conda.anaconda.org/soft-matter/linux-64/tifffile-0.3.1-np19py27_0.tar.bz2
as far as I can see, I did nothing to cause that, and it looks like a transient network error.  Is there any way to force Travis to try again?

That's a good write up, I only stumbled twice:
1) After changing remotes your have to fetch them `git fetch soft-matter`
2) When rebasing to squash commits, I had to pick some, delete some others, squash the third.  Before I deleted my old fork and restarted, I had a merge conflict that was roadblocking me.  Is there a git command that says `I want the current state of the working directory to be the single commit, and ignore how I got here`?

tacaswell:
Sorry about that, wrote that without checking.

git reset --soft sm/master

Will move your current branch to point at the same commit as sm/master and
the diff should show the differences between your current working tree and
the tree on sm/master. You can then commit those differences on to your
feature branch.

caspervdw:
Am I right that #218 fully incorporates this PR? Can this be closed?

nkeim:
That's correct. Closing. @caspervdw thanks for looking at this!

PR #217, opened 2015-12-17
Try 2 on pull request for Support for non-monochrome and packed-bits images in NorpixSeq,

dmopalmer:
but only with the as_raw flag to return a uint8 array.
The user must interpret the raw image correctly, based on the
image_format header entry to determine how to convert
pixel values to RGB, etc.

set_process_func() exposes _validate_process_func() so the user
may open the sequence, examine the header, and then install the process_fun.

This replaces pull-request #216


dmopalmer:
Try 2 didn't work.  Use Try 3 #218

PR #218, opened 2015-12-17
Try 3: Support for non-monochrome and packed-bits images in NorpixSeq,

dmopalmer:
but only with the as_raw flag to return a uint8 array.
The user must interpret the raw image correctly, based on the
image_format header entry to determine how to convert
pixel values to RGB, etc.

set_process_func() exposes _validate_process_func() so the user
may open the sequence, examine the header, and then install the process_func.

### DIFF
--- a/pims/norpix_reader.py
+++ b/pims/norpix_reader.py
@@ -62,8 +62,14 @@ class NorpixSeq(FramesSequence):
     This is the native format of StreamPix software, owned by NorPix Inc.
     The format is described in the StreamPix documentation.
 
-    Currently supports uncompressed files only, either
-    uint8/16/32 monochrome or as_raw color or monochrome.
+    Currently only supports uncompressed files (with the same number
+    of bytes for each frame).
+
+    Only unsigned 8, 16 or 32 bit monochrome files are directly supported.
+    Other color and monochrome pixel formats can be handled by setting
+    the as_raw flag which results in frames with unsigned 8-bit data type.
+    and the correct number of rows.  (If a raw frame size in bytes is not
+    evenly divisible by the number of rows, a 1-D array is returned).
 
     Nominally thread-safe.
 
@@ -92,7 +98,7 @@ class NorpixSeq(FramesSequence):
                        'get_time_float', 'filename', 'width', 'height',
                        'frame_rate']
 
-    def __init__(self, filename, process_func=None, dtype=None, as_grey=False, as_raw = False):
+    def __init__(self, filename, process_func=None, dtype=None, as_grey=False, as_raw=False):
         super(NorpixSeq, self).__init__()
         self._file = open(filename, 'rb')
         self._filename = filename
--- a/pims/norpix_reader.py
+++ b/pims/norpix_reader.py
@@ -23,6 +23,7 @@ __all__ = ['NorpixSeq',]
 DWORD = 'L'
 LONG = 'l'
 DOUBLE = 'd'
+USHORT = 'H'
 
 HEADER_FIELDS = [
     ('magic', DWORD),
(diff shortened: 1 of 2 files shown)


nkeim:
This is great! I restarted the Travis build and you were right about the transient error. That's a relief!

There's enough distinct code in the `as_raw` path that it really needs an automated test. At the very least to make sure that we can add and use a processing function after opening the file. pims was never designed for 1D "images" it seems like if there's not a test for it, someone might come along and break that. 

I would be happy to write a test, but I probably can't get to it for at least a couple weeks.

pims/norpix_reader.py
```diff
@@ -74,7 +92,7 @@ def class_exts(cls):
                        'get_time_float', 'filename', 'width', 'height',
                        'frame_rate']
 
-    def __init__(self, filename, process_func=None, dtype=None, as_grey=False):
+    def __init__(self, filename, process_func=None, dtype=None, as_grey=False, as_raw = False):
```
nkeim: PEP8 (and consistency): There should be no spaces in `as_raw=False`.

pims/norpix_reader.py
```diff
@@ -50,7 +62,8 @@ class NorpixSeq(FramesSequence):
     This is the native format of StreamPix software, owned by NorPix Inc.
     The format is described in the StreamPix documentation.
 
-    Currently supports uncompressed 8-bit monochrome files only.
+    Currently supports uncompressed files only, either
+    uint8/16/32 monochrome or as_raw color or monochrome.
```
nkeim: The "uint" reference could be confusing for casual users, since StreamPix does not use C or numpy names for datatypes. How about "8, 16, or 32 bits per pixel"?

Also, the purpose of `as_raw` might not be obvious from this docstring — that it is essentially a workaround that allows the user to do her own processing of the frame data. It would be good to elaborate on that a bit, to avoid confusion over whether color images are "officially" supported.

pims/norpix_reader.py
```diff
@@ -150,12 +183,17 @@ def _verify_frame_no(self, i):
         if i >= self._image_count or i < 0:
             raise ValueError("Frame number is out of range: " + str(i))
 
+    def set_process_func(self, process_func):
```
nkeim: I think that `pims.pipeline` (derived from Slicerator) already provides equivalent functionality. Could users just use that instead?

dmopalmer: This was an attempt to use the mechanisms that already exist in PIMS.

Since the process_func can be passed to __init__, but you generally need to read the header before you know which process_func to use, the alternative was to open the file, get the header information, close the file, and re-open it with the appropriate process_func.

tacaswell: We are in the process of ripping that functionality back out to be replaced by `pipeline`, sorry for the poor communication on that.

nkeim: Yes, sorry my original comment lacked that context. Basically, in the next major revision of pims, all the `process_func` stuff will be gone. Instead one can/will use something like

```python
@pims.pipeline
def read_as_color(image):
    ...
    return color_image

if is_color(seq_file_opened_as_raw):
    seq = read_as_color(seq_file_opened_as_raw)
elif is_16bit(seq_file_opened_as_raw):
    ...
```

to make a "virtual" sequence object with the desired interpretation.

`pipeline` is already working. So if `set_process_func` is in this PR, I bet that within a few months it will be torn out. But my comment is more of an advisory than a requirement for this PR; someone else can make that change as we update all the other reader classes.

dmopalmer:
OK, I have fixed the PEP8, improved the docstring, and did the pull request for the unit test.

I am now at 4 commits, and once again `git rebase -i` is not my friend.
```
% git remote -v
dmopalmer	https://github.com/dmopalmer/pims.git (fetch)
dmopalmer	https://github.com/dmopalmer/pims.git (push)
soft-matter	https://github.com/soft-matter/pims.git (fetch)
soft-matter	https://github.com/soft-matter/pims.git (push)
% git status
On branch enh_color_norpix
Your branch is up-to-date with 'dmopalmer/enh_color_norpix'.
nothing to commit, working directory clean
% git rebase -i soft-matter/master
fatal: Needed a single revision
invalid upstream soft-matter/master 
```

My guess is that this is because, in the mean time, soft-matter/master has had 17 more commits.  Trying to do a pull of soft-matter/master 

So what do I do now?  (As in 'what mystic incantation do I type?`)

nkeim:
Actually, you're in luck! It looks like none of these changes conflicts with anyone else's, so the merge can proceed automatically as-is.

Thanks again for this enhancement, @dmopalmer . Hopefully we'll have direct support for these variant formats, eventually.

PR #219, opened 2015-12-23
cine: pass native str to struct.Struct

valhallasw:
from __future__ import unicode_literals forces all string literals
to be unicode, but struct.Struct needs a native str. This breaks
loading cine files under Python 2.

This patch creates a new function to create a Struct object, allowing
for either str or unicode to be passed. This also allows us to move
the endianness marker from the main code.

### DIFF
--- a/pims/cine.py
+++ b/pims/cine.py
@@ -26,6 +26,10 @@ import hashlib
 __all__ = ('Cine', )
 
 
+def _build_struct(dtype):
+    return struct.Struct(str("<" + dtype))
+
+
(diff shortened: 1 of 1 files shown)


tacaswell:
Thanks!

I assume that if you found this bug you are using this code?

valhallasw:
Yes, but not very actively. A quick check showed pims reads the file (from a very old Phantom v7.2 camera) without issues, but I haven't done any large-scale analysis. A colleague of mine is currently using a hacked-together Struct-based code, and I will suggest them to switch to pims. :-)

PR #220, opened 2015-12-24
ENH Viewer

caspervdw:
This adds a Viewer based on [`skimage.viewer.CollectionViewer`](http://scikit-image.org/docs/dev/api/skimage.viewer.html). This is useful because a large part of my (an probably many others) workflow is just looking at the images and checking out the effect of different parameters on the image. This kind of work is much more optimal inside for example ImageJ, because you get direct feedback. However, ImageJ is not as fast as PIMS in loading images *and* it doesn't have my favourite particle tracking tool... hence this viewer.

To show the viewer:

    reader = pims.Viewer().show()

Then use the `Open` menu option to open and view a video. When the window is closed, the reader is returned.

You can also display an already opened reader:

    pims.Viewer(pims.open(filename)).show()

### Pipelines
I modified the skimage code for plugins so that it can add pipelines: functions that lazily modify a reader. The nice thing about the Viewer is that you can see the effect of a pipeline directly using sliders to modify the parameters of the pipeline. Because this went pretty far in the code of both the viewer and the plugins, I decided to reimplement the whole `skimage.viewer` class and not subclass it.

Full example:

    import numpy as np
    from pims.viewer import Viewer, ViewerPipeline, Slider

    def add_noise(img, noise_level):
        return img + np.random.random(img.shape) * noise_level

    AddNoise = ViewerPipeline(add_noise) + Slider('noise_level', 0, 100, 0)
    viewer = Viewer()
    viewer += AddNoise
    original, noise_added = viewer.show()

Now the original reader is returned next to a reader that has a pipeline wrapped around.

### 3D rendering
Additonally (because I personally look mainly at 3D images), it wraps the 3D OpenGL viewer from `pyqtgraph` so 3D images can be rendered 3D (something I really like :-D )

### N dimensional files
The viewer expects a `FramesSequenceND`. Because `FramesSequenceND` is aware of its dimensions, it is easy to impelement sliders that makes you able to navigate the reader (for instance in `z`, `c`, `t`). When a `FramesSequence` is used, it is wrapped inside a helper class that makes it behave like a `FramesSequenceND`.

### To do
Many things, but to start:
* better docstrings (I'll say it myself)
* debugging on systems with different python and PyQt/PySide versions
* drag&drop files to open them
* opening lists of files (coupled to `ImageSequence`)
* allow pipelines that change the shape of the reader
* allow plotting pipelines for instance to wrap `trackpy.annotate`


pims/api.py
```diff
@@ -20,6 +20,8 @@
 from pims.tiff_stack import TiffStack_tifffile  # noqa
 from .spe_stack import SpeStack
 
+from pims.viewer import Viewer, ViewerPipeline
```
tacaswell: :-1:  on importing this in the api

caspervdw: Agreed, I removed the import

caspervdw: @tacaswell Or did you mean, put this in a separate repo altogether?

tacaswell:
Just restarting the tests seems to work.  What ever it is it seems to be transient...

caspervdw:
I'll close this PR, because we should decide first if something like this should be in pims. For now I'll be putting it on a seperate repo.

PR #224, opened 2016-01-21
Update bioformats version to 5.1.7

caspervdw:
Travis couldn't download loci_tools.jar. The URL was still right, probably a temporary error. But it made me update bioformats. All tests pass locally.

It will probably fix the travis build when merged.


danielballan:
Yep, Travis is happy again.

PR #227, opened 2016-02-13
PERF ND reader flexibility

caspervdw:
This enables FramesSequenceND readers to use the frame dimensions of the underlying reader in a smart way. For instance: currently when an RGB frame is loaded from a movie, the frame is actually read  3 times, each time only using 1 channel.

Subclassed FramesSequenceND readers have to register their `get_frame` methods, using the axes names:

    self._register_get_frame(self.get_frame_2D, 'yxc')

When `bundle_axes` is set, the reader wraps the `get_frame_2D` method with a `transpose`, `iterate` or `drop` operation so that the correct axes are returned. It always tries to minimize IO traffic: several calls to `get_frame_2D` is referred over reading too much data.

Any combination between `get_frame` axes and `bundle_axes` is possible (and tested).

This enhancement paves the way to make the readers actually lazy in all dimensions. Cropping could be pushed down to reader level, when the underlying reader supports reading pixel lines or single pixels.


caspervdw:
Does anybody have time to look into this PR? The performance increase is great for my purposes and I would like to see it in master.

danielballan:
> This enhancement paves the way to make the readers actually lazy in all dimensions. Cropping could be pushed down to reader level, when the underlying reader supports reading pixel lines or single pixels.

That's very exciting and in line with some things @tacaswell and I are interested in here as well. @vivekthampy might also be excited about this.

I gave this a quick read -- looks fine to me. Merge?

caspervdw:
Merge is fine by me! I am happy to collaborate on extensions or generalizations on this, let me know. I leave the merge to you or @tacaswell

pims/base_frames.py
```diff
@@ -348,11 +479,18 @@ class FramesSequenceND(FramesSequence):
     >>> frames.default_coords['m'] = 3
     >>> frames[5]  # returns Frame at T=5, M=3 with shape (2, 10, 64, 64)
     """
+    def _register_get_frame(self, method, axes):
+        axes = tuple([a for a in axes])
+        if not hasattr(self, '_get_frame_dict'):
+            self._get_frame_dict = dict()
```
tacaswell: I would suggest putting this initialization in `FrameSequenceND.__init__` and requrining sub-classes to call `super()` before doing anything else.   Likely will have to do this test + warn for awhile to maintain back-compat.

tacaswell:
Just one picky comment

caspervdw:
@tacaswell I pushed the changes that you suggested, ready to go

PR #228, opened 2016-02-17
Move docs to soft-matter.github.io

caspervdw:
There was a subtle issue with the documentation: while the docs were configured at http://soft-matter.github.io/ , http://soft-matter.github.io/pims still directed to the gh-pages branch of pims.

I solved this by moving all the docs to http://soft-matter.github.io/docs . Because the 'docs' repository does not exist, this folder is accessible. Now pims docs are at http://soft-matter.github.io/docs/pims and trackpy docs at http://soft-matter.github.io/docs/trackpy .

The pims gh-pages branch should have a redirect to these locations, hence this PR. It basically deletes everything on this branch and adds an index.html that redirects to http://soft-matter.github.io/docs/pims.


caspervdw:
Ok, different approach: as trackpy doesn't have the gh-pages branch, that apparently also solves the problem. I am going to delete the gh-pages branch of pims.

PR #229, opened 2016-03-30
Add a metadata example to the bioformats doc

videan42:
Add an example using the MetadataRetrieve class to access physical pixel dimensions from the image metadata, based on the examples in the openmicroscopy.org docs.

I was reading through the docs trying to figure out how to access the pixel spacing for some light microscopy images, and I couldn't see an example, so I made a quick one. Tested on some Leica Image files I had laying around, but I know the metadata varies from image type to image type, so no promises it works on everything.

It also might be handy to have a way of programmatically listing which metadata tags are available. Right now it's visible in the repr, but you can't actually access it from python.

### DIFF
--- a/doc/source/bioformats.rst
+++ b/doc/source/bioformats.rst
@@ -27,3 +27,26 @@ download the binary from `Christoph Gohlke's website <http://www.lfd.uci.edu/~go
 On first use of pims.Bioformats(filename), the required java library
 :file:`loci_tools.jar` will be automatically downloaded from
 `openmicroscopy.org <http://downloads.openmicroscopy.org/bio-formats/>`__.
+
+Metadata
+--------
+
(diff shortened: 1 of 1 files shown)


caspervdw:
@videan42 Thanks for the PR! It's looking good, I am going to figure out what is going on with Travis but that is unrelated to this doc change.

Could you open an issue for the listing of available metadata tags? I remember I worked on that before, it could be that the function is already in there but just not documented. As this reader is rather new, there is certainly room for improvement so any other issues (or PRs) are welcome!

videan42:
I rebased on the corrected travis.yml, hopefully without breaking things. Let me know if you want any of the text changed or if there are other things I could add.

caspervdw:
Looks good to merge, I'll just wait for #231 to go in so that this change is automatically built into the docs.

caspervdw:
There it goes! Let's hope the automatic doc build does its work...

http://soft-matter.github.io/pims/dev/bioformats.html

PR #230, opened 2016-03-30
Expose the metadata attribute names

videan42:
As requested in #229, add an attribute to MetadataRetrieve to programmatically list the available field names. The accessible fields were calculated during the call to repr, so I just moved that calculation to the bottom of init and assigned it to `self.fields`. Let me know what you'd prefer it to be called.

I added a test that verifies all the expected fields that appear on one of the test images. There's a lot of attributes, so let me know if there's a better image to work with, or if there's a better test than exhaustively listing them all out. Frankly, I only care about a few of the fields, but I guess it's nice to know that they're all there if you need them?

### DIFF
--- a/pims/bioformats.py
+++ b/pims/bioformats.py
@@ -150,6 +150,8 @@ class MetadataRetrieve(object):
                 except ValueError:
                     return field
 
+        self.fields = []
+
         for name in dir(md):
             if (name[:3] != 'get') or (name in ['getRoot', 'getClass']):
                 continue
@@ -168,12 +170,11 @@ class MetadataRetrieve(object):
                                    name + ' wrapped\nby JPype and an '
                                    'additional automatic typeconversion.\n\n')
                     setattr(self, name[3:], fnw)
+                    self.fields.append(name[3:])
                     continue
                 except:
                     # function is not supported by this specific reader
                     pass
-
-        self.fields = list(filter(lambda x: x[:2] != '__', dir(self)))
 
     def __repr__(self):
         return '<MetadataRetrieve> Available loci.formats.meta.' + \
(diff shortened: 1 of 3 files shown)


tacaswell:
can you add `nomkl` to the enviroment created in `.travis`?  I think the 2.7 failures are due to incompatibility between numpy, mkl, and tifffile.

pims/bioformats.py
```diff
@@ -173,10 +173,11 @@ def fnw(fn1=fn, naame=name, paramcount=paramcount):
                     # function is not supported by this specific reader
                     pass
 
+        self.fields = list(filter(lambda x: x[:2] != '__', dir(self)))
```
caspervdw: I think it would make more sense if this is put in the for loop above, at line 171: `self.fields.append(name[3:])`

pims/tests/test_bioformats.py
```diff
@@ -541,6 +541,37 @@ def test_metadata_raw(self):
         assert_equal(metadata['CH2ChannelDyeName'], '5-FAM/pH 9.0')
         assert_equal(metadata['dCalibration'], '0.16780898323268245')
 
+    def test_metadata_tags(self):
+        self.v = self.klass(self.filename, meta=True)
+        fields = self.v.metadata.fields
+        expected_fields = {
```
caspervdw: I think this test is easily broken with different Bioformats versions.
Maybe you can only test a few? Something like `assert 'PixelsPhysicalSizeX' in fields`

videan42:
Made the suggested changes, thanks for the fast review. I'm also stoked that I didn't break Travis...

caspervdw:
Looking good, merging. Thanks for the contribution!

PR #231, opened 2016-04-02
BLD Travis builds the docs on merge in master

caspervdw:
This should automatically build the docs to the dev channel. It only works when travis builds the master branch. The code is based on the matplotlib build.

I tested it on https://travis-ci.org/caspervdw/pims/jobs/120310757, but the code is different as I pushed to a branch on the same repo, while in pims we have the docs on a seperate repo.

I propose that we first merge this and see if Travis builds nicely, then add the `GH_TOKEN` to actually make the push work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,13 @@ sudo: false
 matrix:
   include:
     - python: "2.7"
-      env: DEPS="nomkl numpy=1.9*" DEPSSM="tifffile"
+      env: DEPS="nomkl numpy=1.9*" DEPSSM="tifffile" BUILD_DOCS=false
     - python: "2.7"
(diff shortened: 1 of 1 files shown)


danielballan:
Great!

danielballan:
We can refine the git commands a bit -- there's no need to keep the git history of these build products, so I would propose blowing it away on each build -- but I'm just grateful you took the time to get in working.

caspervdw:
The problem with blowing away the git history is the presence of the other docs. But I guess that is solveable.

It seems to work upto the point that a git token is needed. I am not sure if I have access to the travis internals for soft-matter. Could you generate an access token and add it (as a hidden variable named GH_TOKEN) to Travis? See https://github.com/Syntaf/travis-sphinx

PR #233, opened 2016-05-04
ENH FFmpeg importer/exporters via ImageIO/MoviePy

caspervdw:
This was already in #203, then closed because I wanted to rebase all readers on ImageIO. But as I don't see that happening in the near future, I think this could go in in between.

Additions are `ImageIOReader`, `MoviePyReader`, `export_moviepy`. I use them very frequently because they can import/export any file via ffmpeg.

Optional dependencies are `imageio` and `moviepy`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,21 +4,20 @@ sudo: false
 matrix:
   include:
     - python: "2.7"
-      env: DEPS="nomkl numpy=1.9*" DEPSSM="tifffile" BUILD_DOCS=false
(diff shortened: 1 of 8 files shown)


tacaswell:
:+1: from me.

caspervdw:
@danielballan @tacaswell  Could one of you merge this? I've tested it long enough so I am convinced it should go in. Should solve some issues for people who can't get PyAv built (e.g. me, on both windows and fedora).

danielballan:
Yeah, that's fair. I gave it a quick review. Merging.

PR #235, opened 2016-05-11
BLD Make doc push resemble trackpy

caspervdw:


danielballan:
Related: Are you aware of this? https://drdoctr.github.io/doctr/

As far as I can tell, it is would not easily support our model (multiple repos pushing to one org-level page) but maybe it will someday.

caspervdw:
No I didnt know this. But lets not change something that works...

danielballan:
Oh, 100% agree! Haha. If you ever have a reason to do this again, I thought you'd be happy to know that exists.

PR #241, opened 2016-09-01
spe_stack: Make `frame_shape` match frames' `shape`

lschr:
`frame_shape` used to be (width, height), while the individual frames'
`shapes` are (row count, column count), i. e. the other way round.

### DIFF
--- a/doc/source/release_notes.rst
+++ b/doc/source/release_notes.rst
@@ -1,6 +1,11 @@
 Release notes
 =============
 
+v0.3.4
+------
+- API: Swap elements of ``frame_shape`` in ``SpeStack`` to match frames'
+  ``shape``.
+
 v0.3.3
 ------
 - Fix compatibility with Pillow v0.3.0 (PR 204)
--- a/pims/spe_stack.py
+++ b/pims/spe_stack.py
@@ -233,7 +233,7 @@ class SpeStack(FramesSequence):
 
     @property
     def frame_shape(self):
-        return self._width, self._height
+        return self._height, self._width
 
     def __len__(self):
         return self._len


tacaswell:
:+1:   This needs a note about the API change in https://github.com/soft-matter/pims/blob/master/doc/source/release_notes.rst

lschr:
Just added the note.

doc/source/release_notes.rst
```diff
@@ -1,6 +1,10 @@
 Release notes
 =============
 
+v0.3.4
+------
+- API: Make ``frame_shape`` match frames' ``shape`` in ``SpeStack`` (PR 241)
```
tacaswell: Can you also say that the order has flipped?  The target audience for this text is someone who updated and now their code is broken so they may not be in the mood to dig through code / PRs :smiling_imp: .

lschr:
Better?

doc/source/release_notes.rst
nkeim: An alternative would be to make a new property (image_size?) and have frame_size issue a DeprecationWarning.

On Sep 2, 2016, at 4:44 AM, Thomas A Caswell <notifications@github.com<mailto:notifications@github.com>> wrote:


In doc/source/release_notes.rst<https://github.com/soft-matter/pims/pull/241#discussion_r77333191>:

> @@ -1,6 +1,10 @@
>  Release notes
>  =============
>
> +v0.3.4
> +------
> +- API: Make ``frame_shape`` match frames' ``shape`` in ``SpeStack`` (PR 241)


Can you also say that the order has flipped? The target audience for this text is someone who updated and now their code is broken so they may not be in the mood to dig through code / PRs ? .

-

tacaswell:
Thanks @lschr !

Issue #240, opened 2016-08-25
Conda install downgrades numpy to 1.9


caspervdw:
When installing PIMS 0.3.3 over conda (Win64):

```
conda install pims -c soft-matter
```

Conda wants to downgrade many packages, especially numpy goes to 1.9:

```
The following packages will be DOWNGRADED due to dependency conflicts:

matplotlib:   1.5.1-np111py34_0              --> 1.4.3-np19py34_1
numexpr:      2.6.0-np111py34_0              --> 2.4.4-np19py34_0
numpy:        1.11.1-py34_1                  --> 1.9.3-py34_3
scikit-image: 0.12.3-np111py34_1             --> 0.11.3-np19py34_0
scikit-learn: 0.17.1-np111py34_1             --> 0.16.1-np19py34_0
scipy:        0.17.1-np111py34_0             --> 0.16.0-np19py34_0
```

I really don't know anything about the conda-magic. @danielballan do you have a clue how to fix this?

tacaswell:
We need to rebuild pims with a new version of numpy.  The conflict is probably caused by `tifffile`? Try adding the `conda-forge` channel?

danielballan:
Gotta move to conda-forge in general. That's probably a two-day job.

caspervdw:
Tifffile is on conda-forge indeed. It seems pretty straightforward to move our builds to conda-forge (as long as there is no compiling involved). I started with slicerator [over here](https://github.com/caspervdw/staged-recipes). Are you OK with creating the PR (which would test the build)? @tacaswell @nkeim @danielballan do you want to be maintainers?

tacaswell:
You can add me

caspervdw:
https://github.com/conda-forge/staged-recipes/pull/1688 . If it's merged I'll proceed with pims.

caspervdw:
Ok for some reason I entirely missed this:

https://github.com/conda-forge/staged-recipes/pull/665

It's apparently already on

caspervdw:
@ericdill  Thanks for putting pims and slicerator on conda-forge! I proceeded with the trackpy recipe conda-forge/staged-recipes#1690

caspervdw:
It's all online now. I am :+1: on moving fully from the soft-matter channel to conda-forge, also adapting the installation instructions and release checklist.

- tifffile, fftw, pyfftw are on conda-forge
- ffmpeg with x264 is also present (not Win)
- only thing left: PyAv. Should be doable, except for windows

[...]

PR #245, opened 2016-10-12
Update docs and tests for conda-forge

### LINKED ISSUE #242: conda-forge installation instructions
The [installation instructions](http://soft-matter.github.io/pims/v0.3.3/install.html) need to be updated to reflect the new conda-forge packaging, and to recommend Pillow for IPython display (as referenced in https://github.com/conda-forge/pims-feedstock/pull/1#issuecomment-250999894 )

caspervdw:
Also update docs. Closes #242 and #240.

Additonally, I've added  Py 3.5 tests + mentioned imageio and moviepy in the docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,16 @@ matrix:
     - python: "2.7"
       env: DEPS="nomkl numpy=1.9* slicerator libtiff tifffile" BUILD_DOCS=false
     - python: "2.7"
-      env: DEPS="nomkl numpy=1.9* scipy pillow matplotlib scikit-image jinja2 pyav libtiff tifffile jpype1" DEPSPIP="moviepy imageio" BUILD_DOCS=false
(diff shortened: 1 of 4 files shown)


caspervdw:
All set. I have unpinned the numpy version and added Py3.5 tests.

Do we need legacy tests? In `setup.py`, numpy is limited to 1.7. We could, for example, pin numpy to 1.7 in the first test. Some packages are not distributed below np1.10 (for the recent conda-forge additions `av` and `jpype1`), so this limits our possibilities.

danielballan:
Hmm. At this point I would be comfortable not testing farther back than what is available on conda forge.

caspervdw:
Ok, I did numpy==1.8 for the legacy test, for the rest it is all unpinned. Self-merging as it only touches the build and some docs.

danielballan:
:+1:

PR #247, opened 2016-11-13
ENH Add crop pipeline

caspervdw:
This adds a cropping pipeline that correctly changes the `frame_shape` attribute. I decided to keep the API the same as that of `skimage.util.crop`. There are many other pipelines that we could implement (for instance, rewrite `normalize` and `to_rgb` from `pims.display`, but I think it is a good idea to stay close to skimage and where possible, just import the functions instead of redoing it.

A general approach for that would be testing the function on the first frame at `__init__`, recording the changed attributes, and exposing them automatically. That might save us a fair amount of work, and the performance loss would be not too bad. This would be a `pims` functionality, as it would be specific to images (`frame_shape`, `pixel_type`, etc.).

In this case of `crop`, I did end up copying over the code.

As it makes use of https://github.com/soft-matter/slicerator/pull/21 , we need a slicerator release to get the tests working.


danielballan:
I'll put a slicerator release on my todo list for next week unless you can get to it sooner.

caspervdw:
@danielballan  I'll do the release now

caspervdw:
I would like to start adding some more pipelines, but this has to go in first
Do you agree with exposing pipelines with the same API as skimage, importing where possible?

tacaswell:
What is the current state of this?

rbnvrw:
Hi @caspervdw, we'd like to add this very old PR to the new pims release! Do you have time to update the PR or should we do it?

caspervdw:
@rbnvrw I merged master and attempted to fix an incompatibility with numpy

caspervdw:
@rbnvrw  It seems that slicerator has become incompatible with the newest numpy.

nkeim:
Thanks @caspervdw ! I'm not sure the problem is the newest numpy—our "legacy" Travis environment uses the ancient numpy 1.8.2 and still had similar errors as the other environments, such as this:

```
58ERROR: test_on_frame (pims.tests.test_process.TestCrop)
559----------------------------------------------------------------------
560Traceback (most recent call last):
561  File "/home/travis/build/soft-matter/pims/pims/tests/test_process.py", line 44, in test_on_frame
562    assert_equal(self.pipeline(self.rdr[0]), self.first_frame)
563  File "/home/travis/build/soft-matter/pims/pims/tests/test_process.py", line 57, in <lambda>
564    self.pipeline = lambda x: crop(x, ((5, 32-26), (7, 33-27)))
565  File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/slicerator/__init__.py", line 641, in process
566    return cls(*(tuple([a] for a in ancestors) + args), **kwargs)[0]
567  File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/slicerator/__init__.py", line 480, in __getitem__
568    return self._get(indices)
569  File "/home/travis/build/soft-matter/pims/pims/process.py", line 93, in _get
570    return np.array(ar[self._crop_slices], order=self._crop_order,
571TypeError: list indices must be integers, not tuple
```

I may be able to take a closer look myself. Worst case, we can just hold this PR until a minor release.

nkeim:
@caspervdw I think your commits fixed everything, except for a small mistake. The tests pass now. Could you please check that I got this right?

caspervdw:
Thanks for spotting that! Everything is ok now.

nkeim:
Great! Thanks @caspervdw for writing such a useful and logical feature! I'm glad it will get released.

Issue #244, opened 2016-10-12
Plot_to_frame incompatible with mpl >= 1.5


caspervdw:
The use of `plot_to_frame` with matplotlib 1.5.1 and 1.5.3:

```
    pims.plot_to_frame(plt.figure())
```

Produces the following error:

```
    635 
    636     buf.seek(0)
--> 637     image = np.fromstring(buf.read(), dtype='uint8').reshape(*buf_shape)
    638     if close_fig:
    639         plt.close(fig)

ValueError: total size of new array must be unchanged
```

It seems that the figure shape is behaving differently than on 1.4.3 (on which the  issue is not present)

tacaswell:
Do you have `rcParms['savefig.bbox'] == 'tight'` ?

What are the reported shapes?

I can not reproduce this.

caspervdw:
Spot on, indeed the tight bbox was the cause. Apparently I did not have the same set of rc settings for the different mpl versions.

We need a fix to solve this use case:
```
import pims
import matplotlib as mpl
import matplotlib.pyplot as plt
mpl.rcParams.update({'savefig.bbox': 'tight'})
pims.plot_to_frame(plt.figure())
```

danielballan:
Hmm. Is the right approach to check the contents of rcParams inside our function and just fail more informatively?

tacaswell:
Instead of to the raw RGBA buffer, write it as a png to the bytes array and then read the png.

It is a bit more work (as we have the extra encode/decode step), but it takes care of the size meta-data being properly propagated.

caspervdw:
That sounds reasonable, we just need a check to test the performance overhead.

tacaswell:
png encoding is the method we use in webagg/nbagg so I am not be super worried about that.

PR #248, opened 2016-12-01
FIX Plot_to_frame with safefig.bbox == 'tight'

caspervdw:
This solves issue #244 
There is some performance degrade by going via PNG instead of a raw RGBA array, but it certaintly makes this conversion more robust.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -549,7 +549,7 @@ def to_rgb(image, colors=None, normed=True):
 
 
 @contextmanager
-def _fig_size_cntx(fig, fig_size_inches, tight_layout=None):
+def _fig_size_cntx(fig, fig_size_inches, tight_layout):
     """Resize a figure in a context
 
     Parameters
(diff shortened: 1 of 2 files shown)


tacaswell:
Is the performance drop noticeable?

caspervdw:
I estimated 45ms opposed to 32ms for a single plot conversion. Actually, overriding `savefig.bbox` would be a better idea, but passing `bbox=None` to `savefig` did not have an effect so I went for this.

caspervdw:
@tacaswell I now implemented the png buffer only for the case when tight layout is used. The RC param `savefig.bbox` is temporarily adjusted according to the kwarg `bbox_inches`. If this kwarg is `None`, the `fig.get_tight_layout()` is used.

 - if `bbox_inches=='tight'`, a png buffer is used and `savefig.bbox = 'tight'`
 - if `bbox_inches=='standard'`, an rgba buffer is used and `savefig.bbox = 'standard'`
 - if `bbox_inches==None`, `fig.get_tight_layout()` is called to choose from the above options

tacaswell:
I still do not really understand the purpose of this function (I know I have asked this before, but do not remember).

Not convinced that the complexity is worth it for 10ms, but that is a judgement call.

caspervdw:
I found that a plot to RGB ndarray conversion comes in handy sometimes, for instance if I want to export a plot next to a movie.

Additionally, `tp.annotate3d` depends on `plots_to_frame`, as it is necessary for making an inline scrollable stack of plots. But now the mpl notebook backend is around there is probably a much better solution to this.

Issue #221, opened 2015-12-24
Wrap imread for reading 2D images


caspervdw:
If you try to open a JPG with `pims.open` then it will either fail or (if you have Bioformats installed) use the Bioformats reader. We should have a simple reader that reads the basic image types very fast.

Possibly, this is solved by #203 already. But a better solution might be the same `imread` function as `ImageSequence` uses.

PR #249, opened 2016-12-12
ENH Sequence of ND Readers, Single image reader

caspervdw:
I have added a class `ReaderSequence` that reads a sequence of N-dimensional files, adding an extra axis. Not to be mistaken with `ImageSequenceND`, which reads a sequence of 2-dimensional files into an N-dimensional reader.

For example, `ReaderSequence(filenames, pims.Bioformats, axis_name='p')` reads a series of N-dimensional files using the Bioformats reader, and adds them together on a new axis called `p`. 
It is not possible to bundle this axis, you can only iterate over it.

For the tests I implemented a simple wrapper around `imread` that reads a single image into a lengh-1 reader, both from the 'normal' `FramesSequence` (`ImageReader`) or as an N-dimensional `FramesSequenceND` (`ImageReaderND`). That solves issue #221 .

### DIFF
--- a/pims/api.py
+++ b/pims/api.py
@@ -15,6 +15,7 @@ from warnings import warn
 
 # has to be here for API stuff
 from pims.image_sequence import ImageSequence, ImageSequenceND, ReaderSequence  # noqa
+from pims.image_reader import ImageReader, ImageReaderND  # noqa
 from .cine import Cine  # noqa
 from .norpix_reader import NorpixSeq  # noqa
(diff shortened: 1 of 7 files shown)


danielballan:
This needs a rebase.

caspervdw:
Done

danielballan:
Cool. This is an abstraction we have kind of hacked together at BNL at a higher layer. It's nice to have a way to do this in PIMS.

PR #250, opened 2017-01-03
CLN Remove dtype, process_func and as_gray kwargs

caspervdw:
As discussed previously, this removes the `dtype`, `process_func` and `as_gray` kwargs from all readers. The newer readers already did not have them. Just for reference, the way we now expose these functionalities is through Pipelines:

    @pipeline
    def add_one(frame):
        return frame + 1

    reader_proc = add_one(reader)

Or cast it to a different dtype:

    reader_proc = pipeline(lambda x: x.astype(np.float))(reader)

Or a simple as_gray:

    reader_proc = pipeline(lambda x.sum(axis=2))(reader)

Todo is implement a smarter as_gray function, but that will have to wait until #247 (to avoid a messy rebase)

### DIFF
--- a/doc/source/pipelines.rst
+++ b/doc/source/pipelines.rst
@@ -1,0 +1,115 @@
+Pipelines
+=========
+
+.. ipython:: python
+   :suppress:
+
+   from pims.tests.test_common import save_dummy_png, clean_dummy_png
+   filenames = ['img-{0}.png'.format(i) for i in range(9)]
+   save_dummy_png('.', filenames, (256, 256, 3))
+   from pims import ImageSequence
+   video = ImageSequence('img-*.png')
+
+   def _as_grey(frame):
+       red = frame[:, :, 0]
+       green = frame[:, :, 1]
+       blue = frame[:, :, 2]
+       return 0.2125 * red + 0.7154 * green + 0.0721 * blue
+
+   from slicerator import pipeline
+   as_grey = pipeline(_as_grey)
+
+Videos loaded by pims are (``FramesSequence`` objects) are like lists of numpy
+arrays. Unlike Python lists of arrays they are "lazy", they only load the data
+from the harddrive when it is necessary.
+
+In order to modify a ``FramesSequence``, for instance to convert RGB color
+videos to grayscale, one could load all the video frames in memory and do the
+conversion. THis however costs a lot of time and memory, and for very large
+videos this is just not feasible. To solve this problem, PIMS uses
+so-called pipeline decorators from a sister project called ``slicerator``.
+A pipeline-decorated function is only evaluated when needed, so that the
+underlying video data is only accessed one element at a time.
+
+Conversion to greyscale
+-----------------------
+
+Say we want to convert an RGB video to greyscale. We define a function as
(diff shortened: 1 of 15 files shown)


tacaswell:
👍 This needs some documenation to go with it in the release notes.

Thanks for taking on this work!

caspervdw:
@tacaswell I wrote a docpage on pipelines, see last commit

caspervdw:
@tacaswell @danielballan  Rebased and ready to go if Travis passes. Let's start the next version by cleaning!

doc/source/pipelines.rst
```diff
@@ -0,0 +1,115 @@
...
+   def _as_grey(frame):
+       red = frame[:, :, 0]
+       green = frame[:, :, 1]
+       blue = frame[:, :, 2]
+       return 0.2125 * red + 0.7154 * green + 0.0721 * blue
+
+   from slicerator import pipeline
+   as_grey = pipeline(_as_grey)
+
+Videos loaded by pims are (``FramesSequence`` objects) are like lists of numpy
```
nkeim: Remove the first "are"

doc/source/pipelines.rst
```diff
@@ -0,0 +1,115 @@
...
+arrays. Unlike Python lists of arrays they are "lazy", they only load the data
+from the harddrive when it is necessary.
+
+In order to modify a ``FramesSequence``, for instance to convert RGB color
+videos to grayscale, one could load all the video frames in memory and do the
+conversion. THis however costs a lot of time and memory, and for very large
```
nkeim: Fix case of "THis"

doc/source/pipelines.rst
```diff
@@ -0,0 +1,115 @@
...
+       blue = frame[:, :, 2]
+       return 0.2125 * red + 0.7154 * green + 0.0721 * blue
+
+
+The behavior of ``as_grey`` is unchanged if it is used on a single frame:
+
+.. ipython:: python
+
+   frame = video[0]
+   print(frame.shape)   # the shape of the example video is RGB
```
nkeim: "video" -> "frame"

doc/source/pipelines.rst
```diff
@@ -0,0 +1,115 @@
...
+   processed_frame = as_grey(video[0])
+   print(processed_frame.shape)  # the converted frame is indeed greyscale
+
+
+However, the ``@pipeline`` decorator enables lazy evaluation of full videos:
```
nkeim: I think you want to make it explicit that the modified video can be used exactly as you would use the original pims object. In most cases, it should now look as though you are accessing a grayscale video file, even though the file on disk is still in color.

doc/source/pipelines.rst
caspervdw: I really mean the shape of the frame here, not the video (there it would be video.frame_shape)

caspervdw: Oh wait you mean the next line, sorry

caspervdw:
@nkeim thanks for the comments! This should go in before I would have to do an ugly rebase again.

nkeim:
Great! There may be further opportunities to make this transition painless for users; we should look for them as we approach a release.

PR #254, opened 2017-02-24
Better support for numpy interface.

danielballan:
This is a step toward playing better with dask.

### DIFF
--- a/pims/base_frames.py
+++ b/pims/base_frames.py
@@ -33,6 +33,12 @@ class FramesStream(with_metaclass(ABCMeta, object)):
         """Returns a numpy.dtype for the data type of the pixel values"""
         pass
 
+    @property
+    def dtype(self):
+        # The choice of using the separate name `pixel_type` was historical
+        # and needlessly made PIMS objects look less like numpy arrays.
+        return self.pixel_type
+
     @abstractproperty
     def frame_shape(self):
         """Returns the shape of a single frame as a tuple ex (10, 12)"""
@@ -153,6 +159,10 @@ class FramesSequence(FramesStream):
 
     def __iter__(self):
         return iter(self[:])
+
+    @property
+    def shape(self):
+        return (len(self), *self.frame_shape)
 
     @abstractmethod
     def __len__(self):


nkeim:
:+1: from the user's perspective, too—fewer attribute names to learn.

caspervdw:
Me too! Travis is failing only because of a Py2 incompatibility with the `*self.frame_shape`.

Good to add the appropriate test to `test_common._image_single.test_shape`. I wonder wether it works directly with `FramesSequenceND`?

jakirkham:
Was just starting to look for something like what this library does. Namely the fast lazy loading of big TIFFs while working as nicely as `tifffile` does, which it does quite nicely I might add. 👍

Was interested in getting it working with Dask as well and stumbled on this PR. Let me know if there is anything I can do here to help.

pims/base_frames.py
```diff
@@ -154,6 +160,10 @@ def __getitem__(self, key):
     def __iter__(self):
         return iter(self[:])
 
+    @property
+    def shape(self):
+        return (len(self), *self.frame_shape)
```
jakirkham: Probably can just change this to `(len(self),) + self.frame_shape`.  Though @caspervdw has said as much.

pims/base_frames.py
```diff
@@ -33,6 +33,12 @@ def pixel_type(self):
         """Returns a numpy.dtype for the data type of the pixel values"""
         pass
 
+    @property
+    def dtype(self):
+        # The choice of using the separate name `pixel_type` was historical
+        # and needlessly made PIMS objects look less like numpy arrays.
+        return self.pixel_type
```
jakirkham: Should this be passed through `numpy.dtype` for good measure. Have seen some instances where the type returned by `pixel_type` currently is not a `dtype`, but a `str` representing the `dtype`.

jakirkham:
If anyone is interested, I put together a little piece of code that uses `pims` with `dask` to construct a Dask Array of the image data. Would note this has been hacked together and not at all tested for performance.

ref: https://github.com/jakirkham/pims_dask_exp

pims/base_frames.py
danielballan: Interesting point. Maybe we actually want `pixel_type` and `dtype` to be allowed to be different things in some cases?

danielballan:
Interesting work there. That could lead to a very large graph for a deep image stack, but it's a reasonable approach for medium-sized stacks (large enough to be worth parallelizing, small enough to represent each frame as a node in the graph).

jakirkham:
That's definitely true. I see that code snippet as more of a POC than a battle-tested implementation. Though, as you note, maybe still useful.

Initially I was unsure of how to get PIMS to give me multiple frames in a single array. Seems that passing a sliced image file to any NumPy array construction function just works, which is great. So have updated that code snippet to include an optional setting to specify the number of frames per chunk with a default of 1 frame.

Though have been having similar discussions at Dask about working with large HDF5 files. It seems that when one approaches ~1000 open file handles performance starts to struggle. TBH though would expect one starts to run into many other issues at the same time as you have mentioned with graph size. I'm not sure that this would be of practical concern for us here. Though feel free to let me know otherwise if this is not true.

pims/base_frames.py
caspervdw: I don't think that is necessary, `pixel_type` should return a `np.dtype` by default.

caspervdw:
Do we want to provide a means to convert PIMS `FramesSequence` objects into dask arrays, (something like `reader.to_da_array()`) or do we want to make the `FramesSequence` an (enhanced) subclass of a dask array by default?

My feeling about this is that the `FramesSequenceND` baseclass ([see here](http://soft-matter.github.io/pims/dev/multidimensional.html#make-your-own-multidimensional-reader)) already provides most of the infrastructure to construct a dask array, as it aware of its own dimensions and allows for registering multiple methods that read blocks of data (in most cases, single frames). We might want to start from there.

BTW "normal" `FramesSequence` objects can be wrapped easily into a `FramesSequenceND`, something I implemented for `pimsviewer` over [here](https://github.com/soft-matter/pimsviewer/blob/master/pimsviewer/utils.py#L99).

danielballan:
:+1: for starting from ND and making some `to_dask` method.

jakirkham:
FWIW when I first tried to get a PIMS object into Dask, I tried to use [`imread`]( https://github.com/dask/dask/blob/0.14.1/dask/array/image.py#L19 ). There were a couple issues I encountered.

First the `TiffStack_tifffile` was not picklable. ( https://github.com/soft-matter/pims/issues/265 ) Perhaps other things also suffer from this issue. Not sure if this is a bug or intentional. That said, Dask is going to want to pickle things. So having some kind of solution to this problem will be essential. Second it doesn't seem to be possible to do multiple slices at once. ( https://github.com/soft-matter/pims/issues/262 ) This needs to be handled somehow as Dask wants to treat things like NumPy arrays. The rest of the issues reduce to things that this PR is already addressing. Namely using the standard NumPy interface to allow for a consistent API.

caspervdw:
Those two main problems you describe, are precisely the ones of which I hoped `dask` would solve for us!

Can't we just provide `dask` with a function that reads a certain block of data? (Like `reader.get_frame`)

jakirkham:
So essentially that was what I was trying to do with the code snippet above. FWIW I think that is a perfectly viable solution as well. Am using it that as part of my analysis pipeline currently and it works fine. We are discussing handling HDF5 in Dask with a similar solution actually.

jakirkham:
Maybe @mrocklin has some other thoughts on this that I haven't considered.

FWIW @mrocklin, I think this image library is a better fit for Dask as it lazily loads image files to thus easily allow one to determine shape and dtype without having to read in a single frame. Would be interesting to solve this problem somehow so these two libraries could be used harmoniously.

danielballan:
We have traded some emails about this a (long) while back. Once we've done some basic cleanup on our end we should be able to cook up some interesting examples for the dask community to consider, possibly leading to a better imread in dask and better support for image workflows in general.

jakirkham:
So I decided to take some internal code that was wrapping PIMS with Dask and expose it in a simple package, which I have called [`dask-imread`]( https://github.com/dask-image/dask-imread ). Am planning to make a few finishing touches before putting a package on PyPI and conda-forge. Hoping this will make it easier to interact with image data in Dask. Would definitely appreciate feedback from people interested in this topic.

joshmoore:
Wow, quite the resurrection, @tacaswell.

pims/base_frames.py
tacaswell: done.

tacaswell:
> Wow, quite the resurrection, @tacaswell.

This is why I do not put much stock in "we should close old PRs for the sake of closing old PRs".  If it was a good idea in the past, it is probably a good idea in the future and no reason to trash work for lack of a bit more work to get it finished ;)

pims/base_frames.py
tacaswell: This may break for cases where we do not know their length, but in that case `len(obj)` is already broken so I am not super worried.

jakirkham:
Thanks Thomas! 😄

PR #255, opened 2017-02-26
ENH Timed PyAv reader

caspervdw:
A wrapper around PyAv that interprets `duration`, `average_rate` and the timestamps per frame to give a reader with the indexed based on the timestamp of each frame. Missing frames are inserted as black frames.

Also I dropped the `PIL` dependency of the index-based PyAv reader.

### DIFF
--- a/pims/api.py
+++ b/pims/api.py
@@ -34,13 +34,18 @@ if export is None:
 try:
     import pims.pyav_reader
     if pims.pyav_reader.available():
-        PyAVVideoReader = pims.pyav_reader.PyAVVideoReader
-        Video = PyAVVideoReader
+        PyAVReaderTimed = pims.pyav_reader.PyAVReaderTimed
+        PyAVReaderIndexed = pims.pyav_reader.PyAVReaderIndexed
+        Video = PyAVReaderTimed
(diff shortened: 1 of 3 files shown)


danielballan:
Nice idea. This looks fine to me. Thanks for cleaning up the PIL stuff.

caspervdw:
Ok! I kept the existing reader and renamed it to `PyAvReaderIndexed`. The `PyAvVideoReader` is now an alias for the new reader.

Can be merged when Travis passes!

caspervdw:
On some video files, the `seek` operation raises a segmentation fault, killing Python:

    Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

Do you know if there is a way to catch this? Doing the `seek` from a subprocess?

danielballan:
I think a subprocess is the only way. This is definitely worth reporting upstream. [This recent issue](https://github.com/mikeboers/PyAV/issues/163) _might_ be related.

caspervdw:
It indeed happens for MPEG-2 files. I'll report it upstream. For now I think this can be merged as it should be fixed upstream indeed.

PR #256, opened 2017-03-09
FIX Two fixes for the pyavreader

caspervdw:
Some videos have missing frames, which is mainly caused by codecs that remove frames that are very similar. Therefore the most logical choice seems to return the latest non-empty frame, and not a black frame.

Also, this adds some extra checks when opening the file and the possibility for using video files with multiple video streams.


caspervdw:
As of https://github.com/mikeboers/PyAV/issues/201, the segfaults for MPEG2 files are also out of the way. I pushed another fix to deal with files that don't start at timestamp=0.

danielballan:
Interesting. I didn't know about that feature of some codecs. Good work getting that upstream fix from Mike!

PR #257, opened 2017-03-11
Update and test exporters

caspervdw:
This repairs and tests the PyAV exporter, as well as the moviepy movie exporter. Many tests could be designed still, but it's a start.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -72,6 +72,7 @@ def export_pyav(sequence, filename, rate=30, bitrate=None,
         white values. If the datatype of the images is not 'uint8', this must
         be set to True, as it is by default.
     quality: number or string, optional
+        For 'mpeg4' codec: sets qmin and qmax
         For 'libx264' codec: sets crf. 0 = lossless, 23 = default.
         For 'wmv2' codec: sets fraction of lossless bitrate, 0.01 = default
     options : dictionary, optional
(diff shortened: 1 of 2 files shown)


caspervdw:
This is done, if test pass can be merged.

caspervdw:
@danielballan  Could you check up on this and merge if you agree? Then I will make the release notes and we can release v0.4

pims/display.py
```diff
@@ -33,8 +34,9 @@
 
 
 def export_pyav(sequence, filename, rate=30, bitrate=None,
-                width=None, height=None, codec='mpeg4', format='yuv420p',
-                autoscale=True):
+                width=None, height=None, format=None, codec='mpeg4',
+                pixel_format='yuv420p', autoscale=None, quality=None,
```
danielballan: Would you note the parameter name change `format` -> `pixel_format` in the release notes?

pims/display.py
```diff
@@ -160,25 +202,29 @@ def play(sequence, rate=30, bitrate=None,
...
+            return self._cache[:, :, ::-1]
+        else:
+            return self._cache
 
 
 def export_moviepy(sequence, filename, rate=30, bitrate=None, width=None,
-                   height=None, codec='libx264', format='yuv420p',
-                   autoscale=True, quality=None, verbose=True,
-                   ffmpeg_params=None, rate_range=(16, 32)):
+                   height=None, codec='mpeg4', pixel_format='yuv420p',
```
danielballan: Ditto here.

danielballan:
I now see that you mentioned release notes above. I'll merge this as is and let you address the release notes separately.

danielballan:
...And I see that `format` -> `pixel_format` was documented in #261, just not in this branch (which is where I looked).

PR #258, opened 2017-03-19
spe_stack: Use file size to determine number of frames

lschr:
The number of frames in an SPE file should be recorded in the
file's header. Some software fails to do so correctly. Since SPE files
store data uncompressed, it is possible to calculate the number of frames
from the file size.

A warning is emitted if this doesn't match the number of frames from the
file header.


caspervdw:
Thanks for sharing @lschr !

PR #259, opened 2017-04-03
BLD Append conda-forge channel instead of adding

caspervdw:


danielballan:
Nice. It's good to know about this conda feature. attn @licode

PR #266, opened 2017-04-22
Update bioformats to 5.7.0

caspervdw:
This introduces an issue with ND2 files, at least on my local tests. The rest of the files pass.

### DIFF
--- a/pims/bioformats.py
+++ b/pims/bioformats.py
@@ -41,7 +41,7 @@ def _gen_jar_locations():
     yield os.path.join(os.path.expanduser('~'), '.config', 'pims')
 
 
-def _find_jar(url=None, sha256_checksum=None):
+def _find_jar():
     """
     Finds the location of loci_tools.jar, if necessary download it to a
     writeable location.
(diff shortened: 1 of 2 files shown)


caspervdw:
I investigated this issue, and it is still there for the newest Bioformats 5.7.0.
It seems to appear between 5.3.4 and 5.4.0. The issue occurs at `setId`, the returned error by java is:

`java.lang.IndexOutOfBoundsException: Index: 0, Size: 0`

On 5.7.0 the error is different: now `setId` does not give problems, but the tested ND2 file does not have the right shapes: `rdr.getSizeT()` and `rdr.getSizeZ()` both return `1` while the should return `3` and `10` respectively. 

I will raise issue at the OME mailing list.

caspervdw:
This implements the patch proposed in https://github.com/openmicroscopy/bioformats/issues/2955. Locally the tests pass! @aaron-cognitiv

caspervdw:
Self merging this as test pass. @aaron-cognitiv, please try the updated PIMS code!

PR #268, opened 2017-06-07
Use availability functions to check tests to skip

jakirkham:
Instead of trying to deduce what functionality may or may not be present by using imports, simply use the availability function to determine if the functionality is present or not. This way we can rely on whatever is the best strategy for that module when running the tests for it. Also this gives us the opportunity to verify that the availability functions are working as intended.

### DIFF
--- a/pims/tests/test_common.py
+++ b/pims/tests/test_common.py
@@ -38,16 +38,14 @@ def _skip_if_no_ImageIO():
 
 
 def _skip_if_no_libtiff():
-    try:
-        import libtiff
-    except ImportError:
+    import pims.tiff_stack
+    if not pims.tiff_stack.libtiff_available():
         raise nose.SkipTest('libtiff not installed. Skipping.')
 
 
 def _skip_if_no_tifffile():
-    try:
-        import tifffile
-    except ImportError:
+    import pims.tiff_stack
+    if not pims.tiff_stack.tifffile_available():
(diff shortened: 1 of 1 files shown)


jakirkham:
Thoughts?

pims/tests/test_common.py
```diff
@@ -38,16 +38,14 @@ def _skip_if_no_ImageIO():
 
 
 def _skip_if_no_libtiff():
-    try:
-        import libtiff
-    except ImportError:
+    import pims.tiff_stack
```
caspervdw: Minor thing: I'd prefer `from pims.tiff_stack import libtiff_available()` at the top of the `test` file. Same for the other `_skip` functions.

jakirkham: Sure, just saw other `import`s in functions that were not trying to catch `ImportError`s (e.g. a few lines above) and was doing this to be consistent. Would be totally fine doing it differently.

caspervdw: Right, I am not saying that we are consistent ;-) It's your call, I am fine with merging this.

caspervdw:
Just mentioning #270, we need a proper way to decide what to `import` when.

jakirkham:
Sorry @caspervdw. Forgot to circle back to this one. Thanks for accepting it as is in spite of my oversight. More than happy to work towards better import strategies as raised in the aforementioned issue.

PR #269, opened 2017-06-07
Try scikit-image if tifffile is not available

jakirkham:
In the event that the user does not have `tifffile` installed, but does have `scikit-image` installed, try to import `tifffile` from `scikit-image`'s `external` package. If that still doesn't work, gracefully proceed to marking `tifffile` as missing.

### DIFF
--- a/pims/tiff_stack.py
+++ b/pims/tiff_stack.py
@@ -22,7 +22,10 @@ except ImportError:
 try:
     import tifffile
 except ImportError:
-    tifffile = None
+    try:
+        from skimage.external import tifffile
+    except ImportError:
+        tifffile = None
 
 
 def libtiff_available():


jakirkham:
One could also try falling back to `imageio` after `scikit-image` as they have a vendored copy of [`tifffile`]( https://github.com/imageio/imageio/blob/v2.2.0/imageio/plugins/_tifffile.py ). However it is marked as internal API and doesn't have all of the same functionality that `tifffile` proper or `scikit-image`'s version have like faster code for compression/decompression. That said, it could still be a reasonable third fallback as it is a nice relatively light dependency.

jakirkham:
Thoughts on this?

caspervdw:
I think this is a good idea, but I'll just wait on @danielballan opinion as I really never used this specific reader.

jakirkham:
Ok. FWIW this is a verbatim copy of `tifffile` that is just in `scikit-image` for convenience.

danielballan:
Yep, this is a good idea.

Sorry for being slow, guys. The last two weeks have been bananas!

PR #272, opened 2017-07-04
Add `nd2reader` as a reader

rbnvrw:
[nd2reader](https://github.com/rbnvrw/nd2reader) is a pure-Python package that reads images produced by Nikon NIS Elements 4.0+ microscopes.

Compared to [pims_nd2](https://github.com/soft-matter/pims_nd2), this reader offers some advantages:
- It is not dependent on the Nikon `ND2SDK`, which means the code is fully open-source and can be adapted by anyone.
- It is fully written in Python and does not require any (compiled) platform specific packages
- It can read more metadata than the `ND2SDK` implementation, because the file format has been fully reverse-engineered. For example ROIs are supported.
- The code is (partially) unit tested via CI (Travis) and the code quality is also monitored via CI (Code Climate)
- There is [extensive documentation](http://www.lighthacking.nl/nd2reader/) available

The disadvantages of this package compared to `pims_nd2`are the fact that it can be slightly slower because it is written in Python and that it is not supported by Nikon.

I believe this reader would make a useful addition to the pims project. I will be actively maintaining it, adding new features, fixing bug reports and processing pull requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ PIMS is based on readers by:
    reads and writes many formats)
 * [moviepy](http://zulko.github.io/moviepy) (a Python module that supports
    video editing)
+* [nd2reader](https://github.com/rbnvrw/nd2reader) (a Pure Python module for reading Nikon NIS Elements ND2 images and metadata)
 
 Examples & Documentation
 ------------------------
--- a/pims/api.py
+++ b/pims/api.py
@@ -110,19 +110,19 @@ except (ImportError, IOError):
 
 try:
     from pims_nd2 import ND2_Reader as ND2Reader_SDK
+
+    class ND2_Reader(ND2Reader_SDK):
+        class_priority = 0
+
+        def __init__(self, *args, **kwargs):
+            warn("'ND2_Reader' has been renamed to 'ND2Reader_SDK' and will be"
+                 "removed in future pims versions. "
+                 "Please use the new name, or try out the pure-Python one named "
+                 "`ND2Reader`.")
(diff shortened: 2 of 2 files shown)


caspervdw:
Thanks Ruben! This was indeed on my list to add to pims. Before we merge this, just a naming thing, now we have `ND2_Reader` for the ND2SDK one, and `ND2Reader` for the pure python version.

My opinion on this is to deprecate the name `ND2_Reader` (add deprecation warning) and rename it to `ND2SDKReader`. Then the distinction is clearer.

Also, maybe you can give the class a `class_priority` field, so that it is chosen over the `BioformatsReader` if `pims.open` is used. Just pick a value in between `ND2SDKReader` (20) and `BioformatsReader` (2) because if the Nikon version is installed I'd say it is the preferred one.

rbnvrw:
Thank you Casper.

I agree that the naming of the two packages is a bit confusing now, renaming `ND2_Reader` to `ND2SDKReader` is fine by me.

I have added the `class_priority` of 12 to the master code.

caspervdw:
@rbnvrw  See PR to your branch. Normally I would have write rights if you open a PR, but probably not this time because you used the master branch.

rbnvrw:
Thank you, I merged it.

caspervdw:
@danielballan  or @tacaswell  do you agree with the approach of deprecating `ND2_Reader` in favor of `ND2Reader` or `ND2Reader_SDK`?

tacaswell:
:+1:

caspervdw:
Merged @rbnvrw thanks for the initiative and sorry for the long time it took to merge!

Issue #280, opened 2017-11-09
Incosistent imread() function behavior


charlesreid1:
Hello,

In [`image_sequence.py`](https://github.com/soft-matter/pims/blob/155b01dca6941c4d56ee8f8b075f788ef092b2f5/pims/image_sequence.py#L24-L33) there is a conditional import statement, in which an `imread()` function is imported from either scikit-image, matplotlib, or scipy. However, the behavior of each `imread()` function is different:

* `skimage.io.imread` does not normalize pixel values, so they are integers from 0-255
* `matplotlib.pyplot.imread` does normalize pixel values, so they are floats from 0-1
* `scipy.ndimage.imread` is deprecated in favor of `matplotlib.pyplot.imread` (as printed in the output of calls to scipy's `imread()` function), but it uses integer pixel values 0-255

Also see [this gist](https://gist.github.com/charlesreid1/2087ab15d26a9eb07a13f896b2d7d626).

In issue #121 there was discussion of whether to make scikit-image a dependency, and it sounds like the consensus is no. In light of this, because we cannot count on scikit-image being installed on a user's machine, and because matplotlib is a more lightweight requirement (in keeping with the philosophy of keeping pims lightweight), it might be best to change lines [24-33](https://github.com/soft-matter/pims/blob/155b01dca6941c4d56ee8f8b075f788ef092b2f5/pims/image_sequence.py#L24-L33) to simply be:

```
from matplotlib.pyplot import imread
```

This would affect the output of any script that created an `ImageSequence` object and had `scikit-image` installed (for example, the walkthrough notebooks of trackpy, see issues soft-matter/trackpy-examples#43 and soft-matter/trackpy#455) so it is a potentially breaking change if users are counting on pixel values to be between 0-255. But I believe consistent, predictable behavior across machines is important.

What are your thoughts?

nkeim:
Thanks a ton for shedding some light on this. I agree that this is far from ideal, especially from a scientific reproducibility standpoint.

As a counterargument: I suspect that the vast majority of our users have some kind of comprehensive distribution (e.g. Anaconda) and have skimage, whether they know it or not. Their scripts are all written for non-normalized pixels, and in some cases they care about bit depth and quantization.

What about a one-time warning when pims falls back on matplotlib? We could include a function that suppresses it altogether.

charlesreid1:
[...]

PR #281, opened 2017-11-10
Fix imread() imports

charlesreid1:
This fixes issue #280, with the current approach of three different imread() functions being imported, two of which work differently and one of which is deprecated.

The new behavior is:
* Try to find scikit-image and use `skimage.io.imread()` to read images. This has the preferred imread() functionality.
* If scikit-image is not found, use `matplotlib.pyplot.imread()` and warn the user that this function will import pixel values as floats and not integers. Also tell the user how to turn off this warning message if they get sick of seeing it.
* Remove the attempt to import `imread()` from `scipy.ndimage`, since it is deprecated (also see output in [this gist](https://gist.github.com/charlesreid1/2087ab15d26a9eb07a13f896b2d7d626))

### DIFF
--- a/pims/image_reader.py
+++ b/pims/image_reader.py
@@ -9,22 +9,19 @@ from pims.frame import Frame
 # If scikit-image is not available, use matplotlib (with a warning) instead.
 import warnings
 try:
-    from sklearn.io import imread
-except (ImportError, ModuleNotFoundError):
+    from skimage.io import imread
+except ImportError:
     try:
         from matplotlib.pyplot import imread
         # imread() works differently between scikit-image and matplotlib.
         # We don't require users to have scikit-image, 
         # but if we fall back to matplotlib, make sure the user 
         # is aware of the consequences.
-        skimage_preferred = "pim prefers to read images using scikit-image, but you do not appear to have scikit-image installed. " \
-                            "pim will use matplotlib instead. Note that matplotlib reads pixel values as floats (0-1), not integers (0-255). " \
-                            "\n\n" \
-                            "This might break your script." \
-                            "\n\n" \
-                            "(To ignore this warning, include the line \"warnings.simplefilter(\"ignore\", RuntimeWarning)\" in your script.)"
+        skimage_preferred = "pim's image_reader.py could not find scikit-image. Falling back to matplotlib's imread(), which uses floats instead of integers. This may break your scripts." \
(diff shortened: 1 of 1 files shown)


pims/image_reader.py
```diff
@@ -6,25 +6,30 @@
...
 try:
     from skimage.io import imread
 except ImportError:
     try:
         from matplotlib.pyplot import imread
+        # imread() works differently between scikit-image and matplotlib.
+        # We don't require users to have scikit-image, 
+        # but if we fall back to matplotlib, make sure the user 
+        # is aware of the consequences.
+        skimage_preferred = "pim's image_reader.py could not find scikit-image. Falling back to matplotlib's imread(), which uses floats instead of integers. This may break your scripts." \
```
caspervdw: Could you wrap the lines to 79 characters? And a typo in the fist word: `pim` should be `PIMS`.

caspervdw:
@charlesreid1  Thanks for the contribution! I think this is good to merge, if you address the single (cosmetic) line note I left.

pims/image_reader.py
```diff
@@ -6,25 +6,34 @@
...
+        # We don't require users to have scikit-image, 
+        # but if we fall back to matplotlib, make sure the user 
+        # is aware of the consequences.
+        ski_preferred = "PIMS image_reader.py could not find scikit-image. " \
+                + "Falling back to matplotlib's imread(), which uses floats " \
+                + "instead of integers. This may break your scripts. " \
+                "\n" \
+                "(To ignore this warning, include the line " \ 
+                + "\"warnings.simplefilter(\"ignore\", RuntimeWarning)\" " \
+                + "in your script.)"
```
caspervdw: There appears to be a syntax error somewere overhere. You can use the following:

```
ski_preferred = ("PIMS image_reader.py could not find scikit-image. "
                 "Falling back to matplotlib's imread(), which uses floats "
                 "instead of integers. This may break your scripts. \n"
                 "(To ignore this warning, include the line "
                 '"warnings.simplefilter("ignore", RuntimeWarning)" '
                 "in your script.)")
```

charlesreid1: whoops, same as soft-matter/trackpy/467. Will fix both of these.

caspervdw:
Merging this, thanks @charlesreid1  !

PR #283, opened 2017-12-02
BUG: RGBA movie export

nkeim:
The code to remove the alpha channel before exporting with PyAV or MoviePy had some mistakes. Furthermore, NumPy tends to store RGBA sequence data in a different memory order than RGB, causing problems.

This also makes sure that the exported movie is cleaned up at the end of each test.

### DIFF
--- a/pims/display.py
+++ b/pims/display.py
@@ -149,6 +149,8 @@ def export_pyav(sequence, filename, rate=30, bitrate=None,
                                                       export_rate)
                 stream.bit_rate = int(bitrate)
 
+	# Ensure correct memory layout
+        img = img.astype(img.dtype, order='C', copy=False)
         frame = av.VideoFrame.from_ndarray(img, format=str('rgb24'))
         packet = stream.encode(frame)
         if packet is not None:
@@ -483,9 +485,10 @@ def _to_rgb_uint8(image, autoscale):
         color_axis = shape.index(3)
         image = np.rollaxis(image, color_axis, 3)
     elif image.ndim == 3 and shape.count(4) == 1:
-        # This is an RGBA image. Drop the A values.
+        # This is an RGBA image. Ensure that the color axis is axis 2, and 
+        # drop the A values.
         color_axis = shape.index(4)
-        image = np.rollaxis(image, color_axis, 4)[:, :, :3]
+        image = np.rollaxis(image, color_axis, 3)[:, :, :3]
     elif ndim == 2:
         # Expand into color to satisfy moviepy's expectation
         image = np.repeat(image[:, :, np.newaxis], 3, axis=2)
--- a/pims/display.py
+++ b/pims/display.py
@@ -93,7 +93,6 @@ def export_pyav(sequence, filename, rate=30, bitrate=None,
     export_rate = _normalize_framerate(rate, *rate_range)
     sequence = CachedFrameGenerator(sequence, rate, autoscale)
 
-
     # pyav is picky with unicode strings
     codec = str(codec)
     if format is not None:
@@ -149,7 +148,7 @@ def export_pyav(sequence, filename, rate=30, bitrate=None,
                                                       export_rate)
                 stream.bit_rate = int(bitrate)
 
-	# Ensure correct memory layout
+        # Ensure correct memory layout
         img = img.astype(img.dtype, order='C', copy=False)
         frame = av.VideoFrame.from_ndarray(img, format=str('rgb24'))
         packet = stream.encode(frame)
@@ -324,12 +323,14 @@ def export_moviepy(sequence, filename, rate=30, bitrate=None, width=None,
     clip.write_videofile(filename, export_rate, codec, bitrate, audio=False,
                          verbose=verbose, ffmpeg_params=ffmpeg_params)
 
+
 if av is not None:
     export = export_pyav
 elif VideoClip is not None:
     export = export_moviepy
 else:
     export = None
+
 
 def repr_video(fname, mimetype):
     """Load the video in the file `fname`, with given mimetype,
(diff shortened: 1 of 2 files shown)


pims/display.py
```diff
@@ -481,11 +483,12 @@ def _to_rgb_uint8(image, autoscale):
     if ndim == 3 and shape.count(3) == 1:
         # This is a color image. Ensure that the color axis is axis 2.
         color_axis = shape.index(3)
-        image = np.rollaxis(image, color_axis, 3)
+        image = np.rollaxis(image, color_axis, 2)
```
caspervdw: @nkeim I think this should remain 3. `rollaxis` rolls the axis until the designated one lies before the given number. We want to have the color axis at position `2`, so the last argument should be `3`.

    a = np.empty((3, 100, 100))
    print(np.rollaxis(a, a.shape.index(3), 2).shape)

Gives the incorrect result `(100, 3, 100)`.

pims/display.py
```diff
@@ -481,11 +483,12 @@ def _to_rgb_uint8(image, autoscale):
...
     elif image.ndim == 3 and shape.count(4) == 1:
-        # This is an RGBA image. Drop the A values.
+        # This is an RGBA image. Ensure that the color axis is axis 2, and 
+        # drop the A values.
         color_axis = shape.index(4)
-        image = np.rollaxis(image, color_axis, 4)[:, :, :3]
+        image = np.rollaxis(image, color_axis, 2)[:, :, :3]
```
caspervdw: See previous comment- the third argument should be 3.

pims/tests/test_display.py
```diff
@@ -155,6 +162,25 @@ def test_quality_h264(self):
...
         assert_less(compressed_size, lossless_size)
 
+    def test_rgba_h264(self):
+        """Remove alpha channel."""
+        # Start with smoke test for H.264
+        self.export_func(self.sequence_rgba, self.tempfile, codec='libx264')
+        # Check that RGB channels are preserved
+        self.export_func(self.sequence_rgba, self.tempfile, codec='rawvideo',
+                         pixel_format='bgr24')
+        sequence_rgba_stripped = self.sequence_rgba[:,:,:,:3]
```
caspervdw: Please add spaces after the commas (PEP8)

pims/tests/test_display.py
```diff
@@ -155,6 +162,25 @@ def test_quality_h264(self):
...
+        with pims.open(self.tempfile) as reader:
+            assert_equal(len(reader), self.expected_len)
+            assert_equal(reader.frame_shape, self.expected_shape)
+            for a, b in zip(sequence_rgba_stripped, reader):
+                assert_array_equal(a, b)
+
+    def test_rgba_memorder(self):
+        """RGBA with alpha removed has different memory order."""
+        self.export_func(self.sequence_rgba[:,:,:,:3], self.tempfile,
```
caspervdw: See previous PEP8 style remark

pims/tests/test_display.py
```diff
@@ -110,12 +110,19 @@ class ExportCommon(object):
...
+        self.expected_shape_rgba = (h, w, 4)
         self.expected_len = 20
 
-        self.sequence = np.random.randint(0, 255, size=(self.expected_len,) +
-                                                       self.expected_shape,
-                                          dtype=np.uint8)
+        self.sequence = np.random.randint(0, 255,
+                                          size=(self.expected_len,) +
+                                               self.expected_shape,
+                                          ).astype(np.uint8)
```
caspervdw: Why remove the `dtype` kwarg and use `astype`? Isn't the original version more efficient?

caspervdw:
Thanks @nkeim! I left a couple of line comments and I removed some tabs by pushing to your branch.

pims/tests/test_display.py
nkeim: One of my Python environments has an old version of numpy that doesn't support that kwarg. I'd be OK with reverting this if you'd prefer that we modernize.

pims/display.py
nkeim: You're right! The first line of the `rollaxis()` docstring is a bit misleading…

nkeim:
Thanks, @caspervdw! I just force-pushed a cleaned-up branch that addresses the issues you found.

pims/display.py
```diff
@@ -149,6 +149,8 @@ def export_pyav(sequence, filename, rate=30, bitrate=None,
                                                       export_rate)
                 stream.bit_rate = int(bitrate)
 
+	# Ensure correct memory layout
```
caspervdw: Can you fix the indentation? Sorry for my picky comment.

pims/tests/test_display.py
caspervdw: Ok let’s leave the astype then

nkeim:
Thanks. It looks like tabs were introduced because vim is misconfigured on my compute server. As penance, I ran a PEP8 linter on both files.

pims/tests/test_display.py
```diff
@@ -155,6 +165,25 @@ def test_quality_h264(self):
...
+        sequence_rgba_stripped = self.sequence_rgba[:, :, :, :3]
+        with pims.open(self.tempfile) as reader:
+            assert_equal(len(reader), self.expected_len)
+            assert_equal(reader.frame_shape, self.expected_shape)
+            for a, b in zip(sequence_rgba_stripped, reader):
+                assert_array_equal(a, b)
+
+    def test_rgba_memorder(self):
+        """RGBA with alpha removed has different memory order."""
+        self.export_func(self.sequence_rgba[:, :, :, :3], self.tempfile,
```
caspervdw: Maybe make this more explicit by imposing fortran order? Basically you can copy the `rawvideo export` test with the order change added, that will also test the result.

nkeim: Good idea. The revised test also catches the original bug.

caspervdw:
@nkeim Sorry for the delay, merging this

Issue #296, opened 2018-06-04
PyAVReaderTimed compatibility with PyAV 0.4


alex-amvdor:
Hello,
I am attempting to use trackpy to analyze a .avi video, and as such I need pims.Video to use the video. However, when I attempt to use it, I get an error message that I can't seem to figure out, and it doesn't seem like anyone else on the internet is getting this error message either. 

Here is the code I am running:
`v = pims.Video('C:\\Users\\alexa\\Documents\\College\\Physics\\Physics 25L\\Boltzmann_video_2.avi')`

Here is the complete traceback:
```
Traceback (most recent call last):

  File "<ipython-input-8-18236c661fe1>", line 1, in <module>
    v = pims.Video('C:\\Users\\alexa\\Documents\\College\\Physics\\Physics 25L\\Boltzmann_video_2.avi')

  File "C:\Users\alexa\Anaconda3\lib\site-packages\pims\pyav_reader.py", line 139, in __init__
    frame = next(_gen_frames(demuxer, self._stream.time_base))

  File "C:\Users\alexa\Anaconda3\lib\site-packages\pims\pyav_reader.py", line 54, in _gen_frames
    for packet in demuxer:

  File "av/container/input.pyx", line 88, in demux

  File "av/container/streams.pyx", line 50, in _flatten

TypeError: list indices must be integers or slices, not av.video.stream.VideoStream
```

Someone over at the trackpy section of github told me to check if this was a PyAv issue or a pims issue using:

```
import av

container = av.open(path_to_video)

for frame in container.decode(video = 0):
      frame.to_image().save('frame-04%d.jpd' % frame.index)
```

Which returned:

`deprecated pixel format used, make sure you did set range correctly`

nkeim:
HI! Sorry for the long delay, and thanks for checking PyAv. The warning that it gave you is just a warning, not an error, so it should have worked. Are there files in your working directory? (That would be the directory you were running the script in, or, if you are using a notebook, it is usually the directory where the notebook is stored. It is not necessarily the directory where the movie is stored.)

lyprince:
I am having this problem also. I had no problems with pims before, but after installing av this error is occurring. Both were installed with conda in python 3.5

nkeim:
[...]

PR #300, opened 2018-06-27
Fixed problem with PyAV 0.4.0 and further code for video reading

Meehai:
Should fix this: https://github.com/soft-matter/pims/issues/296

Hello, I'm new to both these projects (PIMS and PyAV), so if I did something wrong, I'm sorry.

Basically I traced the problem to that code, where it seems that the get(self, *args, **kwargs) function in av/container/streams.pyx in PyAV can receive either:
- A stream, in which case the stream itself is returned
- An index, in which case the nth "stream" from their internal list of streams is returned (member _streams)
- A dictionary, in which case the key must be one of the members of that class (_streams, video, audio, subtitles, data, other) and the value must be a list of indexes (or something else, in which case it's considered it's an int, and converted to a list)

What PIMS (this library) sends is a dictionary, with key _streams and value a Stream, so it tries to call
self._streams[StreamObject], which isn't an index and it fails.

Since that StreamObject is in the _stream list (I guess they should check that), it can be passed simply (without keyword argument, in which case it's treated as a dict in their code), so the 

``
elif isinstance(x, Stream):
      selection.append(x)
``

is taken, and the Stream is simply returned.

Sorry for the big story and if there's a better fix, do that instead :)

Mihai

### DIFF
--- a/pims/pyav_reader.py
+++ b/pims/pyav_reader.py
@@ -132,7 +132,7 @@ class PyAVReaderTimed(FramesSequence):
         self._cache = [None] * cache_size
         self._fast_forward_thresh = fast_forward_thresh
 
-        demuxer = self._container.demux(streams=self._stream)
+        demuxer = self._container.demux(self._stream)
 
         # obtain first frame to get first time point
         # also tests for the presence of timestamps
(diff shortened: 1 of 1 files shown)


caspervdw:
@Meehai  Thanks a lot for your fix.  There are some failing tests still, but these are unrelated (having to do with the pyav export function, and tifffile import). Merging.

Issue #310, opened 2018-10-08
Can I provide a list of addresses to pims.open?


mrocklin:
I have several images on a cloud object store.  I can get normal http routes to these images and I'm happy to see that pims can read these individually.  However, my guess is that normal `glob` processing won't successfully find all of the images on the cloud object store.  I'm happy to provide this list myself, but it looks like I'm unable to past a list of filenames into `pims.open`.

```python
In [1]: import pims

In [2]: url = 'https://pydata.org/images/logo.png'

In [3]: pims.open(url)
Out[3]: 
<Frames>
Length: 1 frames
Frame Shape: 53 x 126
Pixel Datatype: uint8

In [4]: pims.open([url, url, url])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-4b4693a3872c> in <module>()
----> 1 pims.open([url, url, url])

~/Software/anaconda/lib/python3.6/site-packages/pims/api.py in open(sequence, **kwargs)
    147     >>> frame_shape = video.frame_shape # Pixel dimensions of video
    148     """
--> 149     files = glob.glob(sequence)
    150     if len(files) > 1:
    151         # todo: test if ImageSequence can read the image type,

~/Software/anaconda/lib/python3.6/glob.py in glob(pathname, recursive)
     18     zero or more directories and subdirectories.
     19     """
---> 20     return list(iglob(pathname, recursive=recursive))
     21 
     22 def iglob(pathname, *, recursive=False):

~/Software/anaconda/lib/python3.6/glob.py in _iglob(pathname, recursive, dironly)
     38 
     39 def _iglob(pathname, recursive, dironly):
---> 40     dirname, basename = os.path.split(pathname)
     41     if not has_magic(pathname):
     42         assert not dironly

~/Software/anaconda/lib/python3.6/posixpath.py in split(p)
    103     """Split a pathname.  Returns tuple "(head, tail)" where "tail" is
    104     everything after the final slash.  Either part may be empty."""
--> 105     p = os.fspath(p)
    106     sep = _get_sep(p)
    107     i = p.rfind(sep) + 1

TypeError: expected str, bytes or os.PathLike object, not list
```

Note that this behavior is currently advertised in the docstring 

```python
In [5]: pims.open?
Signature: pims.open(sequence, **kwargs)
Docstring:
Read a filename, list of filenames, or directory of image files into an
iterable that returns images as numpy arrays.

Parameters
----------
[...]

PR #316, opened 2019-01-06
Ensure `open` handles `list`s correctly

jakirkham:
Fixes https://github.com/soft-matter/pims/issues/310

The call to `glob` was causing `list`s of files to fail in `open`. This fixes that issue by looping over the `list`s contents and concatenating them.


rbnvrw:
@jakirkham is this PR still up to date?
Did you see the suggestion made in the referenced issue? https://github.com/soft-matter/pims/issues/310#issuecomment-456193488
If you could update the PR, I'll be happy to merge it.

PR #317, opened 2019-01-17
Cine features

baneel:
Hi all,

Here is an update of the Cine class (to read RAW .cine files), and even more than that. I implemented different and relatively independent things, that I found useful for me and that work on a few examples (saved with PCC v3.3).
1. Changed and made uniform and consistent the declaration of data types ([`python`](https://docs.python.org/3.6/library/struct.html) and [`cine`](http://phantomhighspeed-knowledge.force.com/servlet/fileField?id=0BE1N000000kD2i))
2. Updated the SETUP_FIELDS with fields up to 'cine_name'. In particular, the field with crop info was useful to me.
3. Suggested to change lists of fields-value (e.g. bin_name#:value) into a single field, containing a tuple (bin_name: ()).
4. Changed the names of some fields, when they are said to be obsolete: added 'Res_' (for reserved) at the beginning of the key.
5. Added a _clean_setup_dict() function that cleans up the setup dictionary (filter out obsolete field, remove '\x00', etc). I thought it to be easily overwritten by any user, to match individual further use.
6. Added a _remove_trailing_x00(dict) function that remove the trailing blank characters \x00. This one can be improved.
7. Rearranged the order of __init__ to make the Cine class work with .chd files, at lowest order.
8. Wrote the correct computation of the frame rate, based on the image timestamps. Raises a warning when the precision (std/mean) is greater than some value. The frame rate given by the cine file is stored in setup_fields_dict['frame_rate_recorded'].
9. Changed the get_time function accordingly, based on timestamp (minus first image).
10. Added a 'shape' property, to mimic the np.array corresponding to the images series.
11. Changed the name of some methods to hide them on interactive prompt (e.g. unpack -> _unpack)
12. Wrote a save_image_sequence() function to export to hard drive the images. It includes crop and conversion to any image data type via skimage.exposure.rescale_intensity. Not really computationnally efficient, there should be a way to make it faster.

The TODOs that I see for now are in the file.

Thanks for your reviews!
Baptiste


pims/cine.py
```diff
@@ -68,147 +79,204 @@ def _build_struct(dtype):
...
+    ('sensor', UINT32),
+    ('shutter_ns', UINT32),
+    ('edr_shutter_ns', UINT32),
+    ('frame_delay_ns', UINT32),
+    ('im_pos_xacq', UINT32),
+    ('im_pos_yacq', UINT32),
+    ('im_width_acq', UINT32),
+    ('im_height_acq', UINT32),
+    ('description', '4096s'),
+    ('rising_edge', BOOL),
```
nkeim: Are these attributes present in all recent versions of the CINE format? As I note in the comment below, it is not unusual to want to read older experimental movies, so compatibility should not be broken unless it is necessary.

pims/cine.py
```diff
@@ -288,13 +358,52 @@ def __init__(self, filename):
                                                    })
         self.stack_meta_data['trigger_time'] = self.trigger_time
 
+        ### IMAGES
+        # TODO: add support for reading sequence within the same framework, when data
+        # has been saved in another format (.tif, image sequence, etc)
+        if splitext(filename)[-1] in ['.cine', '.cci', '.cin']:
```
nkeim: A different test based on file size or contents would be much better. File extensions can be unreliable, and in general PIMS only uses extensions for the automatic `open()` feature, not within individual readers. At minimum, this test should be case-insensitive.

pims/cine.py
```diff
@@ -288,13 +358,52 @@ def __init__(self, filename):
...
+        setup['f_tone'] = tuple((tone[2*k], tone[2*k+1])\
+                                for k in range(setup['tone_points']))
+        return None
+
+    def _remove_trailing_x00(self, dic):
+        for k, v in dic.items():
+            if isinstance(v, bytes):
+                try:
+                    dic[k] = v.decode('utf8').replace('\x00', '')
+                except:
```
nkeim: This block, and the one below, need to catch a specific exception class or list of classes.

pims/cine.py
```diff
@@ -288,13 +358,52 @@ def __init__(self, filename):
...
+                    pass
+
     @property
     def filename(self):
         return self._filename
 
     @property
     def frame_rate(self):
-        return self.setup_fields_dict['frame_rate']
+        """Actual frame rate, averaged on frame timestamp (Hz)."""
```
nkeim: I think this is a definite improvement! This is the correct behavior in nearly all circumstances, and it is good to have the `frame_rate` attribute available also.

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+        times = np.r_[[t[0].timestamp() + t[1]\
+                       for t in self.frame_time_stamps]]
+        periods = 1/np.diff(times)
+        fps, std = periods.mean(), periods.std()
+        if std/fps > relative_error:
+            warnings.warn('Precision on the frame rate is above {:.2f} %.'\
+                 .format(1e2*relative_error))
+        return fps
+
+    def get_frame_rate(self):
```
nkeim: This is a better name for the function than `get_fps`, but it is not worth breaking the API.

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
-        # which is read above.
-        # See NorpixSeq for a timestamp API that solves this problem.
-        return float(i) / self.frame_rate
-
-    def get_fps(self):
+    def get_time(self, j):
+        """Get the delta time (s) between frames j and 0."""
+        times = [self.frame_time_stamps[k] for k in [0, j]]
+        t0, tj = [t[0].timestamp() + t[1] for t in times]
+        return tj-t0
```
nkeim:
```suggestion
        return tj - t0
```

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+
+    def _compute_frame_rate(self, relative_error=1e-3):
```
nkeim: `max_relative_error` or `relative_error_toler` would be more informative

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+            raised.
+
+        Returns
+        -------
+        fps : float.
+            Actual mean frame rate, based on the frames time stamps.
+        """
+        times = np.r_[[t[0].timestamp() + t[1]\
+                       for t in self.frame_time_stamps]]
+        periods = 1/np.diff(times)
```
nkeim: PEP8
```suggestion
        periods = 1 / np.diff(times)
```

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+        fps, std = periods.mean(), periods.std()
+        if std/fps > relative_error:
```
nkeim: PEP8

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+            warnings.warn('Precision on the frame rate is above {:.2f} %.'\
```
nkeim:
```suggestion
            warnings.warn('Precision on the frame rate is above '
```

pims/cine.py
```diff
@@ -506,16 +621,45 @@ def __len__(self):
...
+                 .format(1e2*relative_error))
```
nkeim:
```suggestion
                 '{:.2f}%.'.format(1e2*relative_error))
```

pims/cine.py
```diff
@@ -22,32 +24,41 @@
 from threading import Lock
 import datetime
 import hashlib
+from os.path import split, join, splitext
+import os
+import subprocess as sbp
+from skimage import io, exposure
```
nkeim: If at all possible, `skimage` should remain an *optional* dependency of PIMS.

pims/cine.py
```diff
@@ -586,6 +730,97 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self == other
 
+    def save_image_sequence(self, base_dir=None, im_ext='.tif', fmt='06d',
```
nkeim: This is a very useful addition, but I don't think it should go here. Only the `crop` option is specific to the CINE format. Everything else would apply to all 2D sequences in PIMS. The ffmpeg export in particular is already found in the `display` module. (It is not well-documented, but since it has been tested and used more widely, it is better to build on that functionality.)

It would be better to add a new module, maybe called `export`. @danielballan , what do you think?

Could `crop`, and the use of `white_level` and `black_level`, be options for the reader class itself?

pims/cine.py
```diff
@@ -586,6 +730,97 @@ def __eq__(self, other):
...
+            take the path of the cine file.
+
+        im_ext : str, optional.
+            Single image extension. Default is tif, other values may lead to 
+            unsuspected behaviour (compression, etc).
+
+        fmt : str, optional.
+            String formatting of image labeling.
+
+        starts_with : int, optional.
```
nkeim: This option seems unnecessary, because of the existing slicing functionality in PIMS.

pims/cine.py
```diff
@@ -586,6 +730,97 @@ def __eq__(self, other):
...
+            First saved image number, default 1.
+
+        method : str, optional.
+            Choose between 'skimage' (default), 'ffmpeg'.
+            See description of formats in Notes.
+
+        crop : bool, optional.
+            Crop the output image series (requires data to be stored in the .cine).
+
+        out_range : str.
```
nkeim: This option also seems unnecessary; it would be much better for this transformation to be a PIMS pipeline so it can be used elsewhere.

pims/cine.py
```diff
@@ -586,6 +730,97 @@ def __eq__(self, other):
...
+        # Init file, folder names
+        base, fname = split(self._filename)
+        if base_dir is not None:
+            base = base_dir
+        prefix, ext = splitext(fname)
+        fol = join(base, prefix)
+        try:
+            os.mkdir(fol)
+        except (FileExistsError):
+            ans = input('Folder {} already exists. Overwrite? (y/n) '.format(fol))
```
nkeim: This has to use a different approach because PIMS is often used for batch processing, not interactively. Perhaps there could be a keyword argument that controls this behavior? One of the choices could be `'ask'`, but I'm not sure that it should be the default.

pims/cine.py
```diff
@@ -586,6 +730,97 @@ def __eq__(self, other):
...
+            # Currently, spans original 12 bits image range over 16 bits.
+            # TODO: implement cropping 
+            warnings.warn('ffmpeg method may lead to undesired behaviour, '\
+                          +'such as spanning 12 bts image over 16 bits. '\
+                          +'Consider using other available method.',
+                          UserWarning)
+            call = 'ffmpeg -i ' + self._filename \
+                   +' '+join(fol, prefix)+'%'+fmt+im_ext
+            sbp.call(call, shell=True)
+        elif method == 'skimage':
```
nkeim: `skimage` does not need to be a dependency. See the beginning of `image_sequence.py` for an example of the fallbacks you can use.

tacaswell:
Good to see you here @baneel !

pims/cine.py
baneel: Yes, from version 637 up to 'description', v. 658 for 'rising_edge' (see doc p. 49).
With the _remove_trailing_x00 function, those parameters would just appear as empty string for older cine files, wouldn't they?

tacaswell:
I gave it a quick read, my biggest concern can be boiled down to back-compatibility (both of reading old files and of making sure old code does not break).

pims/cine.py
```diff
@@ -68,147 +79,204 @@ def _build_struct(dtype):
...
-    ('saturation', INT),
-    ('res_5', BYTE),
-    ('auto_exposure', UINT),
+    ('length', UINT16),
+    ('Res_13', UINT16),
+    ('sig_option', UINT16),
+    ('bin_channels', INT16),
+    ('samples_per_image', UINT8),
+    ('bin_name', ''.join(8*['11s'])),
+    #[('bin_name{:d}'.format(i), '11s') for i in range(8)] + [
```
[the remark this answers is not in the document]
baneel: You understand it correctly, this is a only a suggestion, that consciously violates backwards compatibility. It would be easy, however and with that initial unpacking, to add an option/private method to 'unzip' those vectors and recover the old behaviour (bin_name is concerned, as well as others).

pims/cine.py
```diff
@@ -264,16 +335,15 @@ def __init__(self, filename):
...
-
         self.tagged_blocks = self.read_tagged_blocks()
         self.frame_time_stamps = self.tagged_blocks['image_time_only']
         self.all_exposures = self.tagged_blocks['exposure_only']
         self.stack_meta_data = dict()
         self.stack_meta_data.update(self.bitmapinfo_dict)
         self.stack_meta_data.update({k: self.setup_fields_dict[k]
                                      for k in set(('trig_frame',
-                                                   'gamma',
-                                                   'frame_rate',
```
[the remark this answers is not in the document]
baneel: Actually, none of them is given as obsolete by the documentation...

frame_rate (FrameRate in the doc by VisionResearch) is not obsolete (FrameRate16 is obsolete, see the [`doc by VR`](http://phantomhighspeed-knowledge.force.com/servlet/fileField?id=0BE1N000000kD2i), p. 16). However, the parameter is based on a software input and may not be reliable for every application (e.g. external trig), I found it more useful to compute it directly from the timestamps and store that value in 'frame_rate' (the correct one). The software value is in the new key 'frame_rate_recorded'.

We can make the contrary: give 'frame_rate' the software value and add a new key (say, 'frame_rate_avg') for the average one.

gamma is not obsolete, in the sense that it was not removed by VisionResearch, but it has been replaced by f_gamma (f for float I suppose, see doc pp. 21 and 48). Both could be kept.

pims/cine.py
baneel: Such as testing the content after the image offset? I'll propose some stuff based on that idea.

pims/cine.py
baneel: Ok, let's move it away for now. This will solve the problem of the skimage dependency as well.
Anyway, I had to come back to it, I'm experiencing some trouble with my movies export. I'll work on my own on it and if I arrive to some satisfactory, general result, I'll submit the code, in a separate `export` module.

`crop`, `white_level` and `black_level` are in the `Cine.setup_fields_dict`. They can easily be implemented as options.

pims/cine.py
nkeim: Great! Yes, I think comparing the total file size to the image offset would work well. I am thinking of use cases where people are reading from object stores or other file-like objects where the extension is not always given.

It just occurs to me: isn't a .chd file just a .cine file with zero images? That should not cause any problem for the code and would be easy to detect with the existing `__len__()` implementation.

pims/cine.py
nkeim: OK. Be sure to check out the ffmpeg code in `display.py`; it is done in memory so it is relatively fast.

nkeim:
I just remembered that we do not have any unit tests for `cine.py` at all! This certainly makes it harder to have this discussion. @baneel would you like me to make a very minimal set of tests, based on the ones I wrote for `norpix.py`?

pims/cine.py
baneel: Removed.

pims/cine.py
baneel: The function is removed. I will check `display.py`.

pims/cine.py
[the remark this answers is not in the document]
baneel: Ok. I kept all the previous names, for API compatibility.
I added 2 dict, UPDATED_FIELDS and TO_BE_IGNORED_FIELDS, based on the documentation nomenclature. Anyone can add a filter, based on that.

`frame_rate` stays the same, the computed version becomes `frame_rate_avg`.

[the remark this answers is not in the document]
baneel: See commit ae54e20.

I refrained from coding a warning (for the old get_fps behaviour), would it be relevant?

pims/cine.py
baneel: Yes, please.

I'll produce some unit tests as well.

pims/cine.py
baneel: Same remark as above: `frame_rate`, `get_fps`, `setup_fields_dict['frame_rate']` keep the same behaviour. `frame_rate_avg` and `_compute_frame_rate()` are introduced.

pims/cine.py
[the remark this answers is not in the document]
baneel: Ok, I just rendered the code uniform, back to the previous behavior for those fields.

baneel:
Back to work on that branch.
I made a brief tour on the code, and got back to their original behavior what had to be (I just added a PendingDeprecationWarning in the get_time function).
Let's wrap up what's new:
- self._clean_setup_dict, to clean the dictionary. This function is called at the end of __init__, and could be rewritten (or any public method) to fit user specific purpose.
- self._remove_trailing_x00, so the setup looks a bit better.
- self.get_time_to_trigger, that gives the actual time, based on the timestamp. I added this feature as a frame metadata.
- self.shape: (N, H, W), to mimic np.ndarray image sequence shape (for gray images)
- self.get_frame_rate_avg: compute average frame rate, based on get_time_to_trigger times.
- self.frame_rate_avg: previous function turned into an attribute.

Hope I don't forget anything important.

And now I should run the code on unit tests, and format the ones I get.

pims/cine.py
baneel: Ok,
I wrote a new feature that stop the SETUP_FIELDS at the good location, depending on the software version with which it was recorded.
It works on your cine file, I still have to produce a small one (the ones I planned to share are too big).

pims/cine.py
baneel: Done.
I finally put the function in the module (at the bottom of the file), rather than a class method.

nkeim:
This is excellent! Looking forward to having the unit tests included. Let us know when we should do another review.

baneel:
Well,
I'm afraid we're good! You can go for another (hopefully final) review.
Cheers, and thanks!

nkeim:
It looks like there are some failing tests. There are lots of ways to try the tests yourself, but the simplest way is to just open a terminal, go to the package directory, and run `nosetests`. 

You can also include options that make it run `test_cine.py` only, to save time.

One error I noticed on Travis is that the `collections.abc` module is missing from the oldest Python version we test with. If `abc` is really important to how the new Cine code is written, we could just drop support for that very old version of Python.

baneel:
I'm pretty new to code testing, and I don't fully understand the Travis failing.
But well I've cleaned the imports, especially with the `Iterable` class now directly coming from `collections`, and fix the bugs I spotted.
And `nosetests` runs fine now.

baneel:
After a more careful reading of the logs of the failings builds, I could not find any error directly related to my contribution:
- [for python 2.7](https://travis-ci.org/soft-matter/pims/jobs/516268109): `RuntimeError: imageio.ffmpeg.download() has been deprecated. Use 'pip install imageio-ffmpeg' instead.'`
- [for python 3.6](https://travis-ci.org/soft-matter/pims/jobs/516268111): there is one warning about incompatible `tight_layout`, and the following `TypeError: 'float' object cannot be interpreted as an integer`

Is there a way to fix that in this pull request or should it be the occasion for a new issue or pull request?

caspervdw:
@baneel  I had a quick look at the failing builds, and solved two issues. They are for sure unrelated. I also left some line notes.

pims/cine.py
```diff
@@ -668,3 +887,20 @@ def _sixteen2twelve(b):
         a[k+2:k2:3] = ((b1 & 0x0FF) << 0)
 
     return a
+
+
+def _remove_trailing_x00(dic):
+    """Remove binary null character b'\x00' in dictionary dic entries."""
```
caspervdw: This converts bytes to str as well, right? Maybe good to document that

pims/cine.py
```diff
@@ -668,3 +887,20 @@ def _sixteen2twelve(b):
...
+    for k, v in dic.items():
+        if isinstance(v, bytes):
+            try:
+                dic[k] = v.decode('utf8').replace('\x00', '')
```
caspervdw: You don't need to `decode` if you want to use replace, you can just do `.replace(b'\x00', b'')`

pims/cine.py
baneel: Yes, it does. I updated the docstring.

pims/cine.py
baneel: Ok, I simplified the code, thanks for the tip.

baneel: I realize that my previous solution with `decode` outputs a `str`, which was the desired behavior. The solution with only `.replace` in commits [`e9b0a5f`](https://github.com/soft-matter/pims/pull/317/commits/e9b0a5fde880fe996bafc4b62275c8c107b908f3) and [`87e8e45`](https://github.com/soft-matter/pims/pull/317/commits/87e8e45ebf374d5e36ffae0a836e30d9b4cb8d25) returns a byte. Can we go back to the previous solution?

nkeim: @baneel It sounds like you should go back to your old solution with `decode` if the new one does in fact fail!

baneel: @nkeim, which is what I've committed. Thanks!

baneel:
Sorry about my lag about all that, I had cherry-picked the 2 commits of #319 (15c22aa and 5e2a286 in this PR) prior to work on it and add my own test (I follow the discussion about size in #319).
That created a conflit, which I solved (commented lines which were uncommented).

nkeim:
Excellent! Thank you for this exciting contribution, @baneel , and thank you for your help, @caspervdw  !!

PR #319, opened 2019-01-27
TST: Add basic tests of Cine reader

nkeim:
This adds some long-awaited basic tests for `cine.py`. It includes a "small" (3.2 MB) sample movie created in 2008, to help ensure backward compatibility.

Still needed: A modern Cine file to test against, and more detailed tests of metadata and image reading. I no longer use Phantom cameras so I will leave this to @baneel or another contributor.


caspervdw:
@nkeim  Great to have some unittests for the cine reader, but, 3 MB? It is a bit large for git in my opinion. Is there a way to only include 2 or 3 frames in the file?

nkeim:
@caspervdw I agree! Unfortunately that was the smallest old .cine file I could find. The only possible solutions I can see are:
- If anyone still has a ~12-year-old version of the software we could make a smaller file. 
- Alternatively, we could spin off a `pims-test-files` repo.
- I could open the file in a hex editor and attempt surgery :).

I believe my file also got added to #317 . If we wanted we could wipe out the file and then create an open issue, to test with the old format.

nkeim:
@caspervdw Following up more on your comment on file size for test movies: The PIMS distribution is currently 11.3 MB, of which 7.3 MB is test data, and 3.2 MB is example data. The largest file is `stuck.tif`, at 2.6 MB. The new test CINE movies in #317 (legacy and new formats) would add 4.9 MB. 

The increase in repo size does not seem significant (10%), but the increase in the distribution package size is significant and it will make pims a poorer citizen of the Python ecosystem. Fortunately we can remove the files at any time in the future if we want to slim down the package. I would also note that this is to support an important feature that is integral to pims for (I think) many users.

Given the (clearly) limited time we all have to work on this, and the importance of tests as we add @baneel 's major changes to `cine.py` in #317 , I think we should proceed with merging #317 when it's fully ready. But I am open to persuasion if there is another way out!

caspervdw:
@nkeim Thanks for working on these tests, I agree that we should be more pragmatic at PIMS.

I thought test data wasnt included in distributions?

baneel:
My idea was to create a gently pathological .cine file, with some decimation (in particular to test the features I introduced), and I based its size on the existing test case. I could do smaller if we decide so, without much additional effort.

Please note that the test case I provided was added directly in #317, along with the current cherry-picked commits, which, I think, makes this thread now irrelevant, except for the discussion (and maybe further work on the test cases).

Thanks for all!

nkeim:
@caspervdw You're right; I hadn't appreciated that by default only `.py` files are in the source distribution. Thanks!

PR #320, opened 2019-03-11
Base ImageIOReader on FramesSequenceND

rbnvrw:
Base `ImageIOReader` on `FramesSequenceND`
The `FramesSequenceND` class is more versatile than the `FramesSequence` class.


pims/imageio_reader.py
```diff
@@ -39,6 +46,23 @@ def __init__(self, filename, **kwargs):
         self._shape = first_frame.shape
         self._dtype = first_frame.dtype
 
+        self._setup_axes()
+        self._register_get_frame(self.get_frame, 'tyx')
```
caspervdw: I think it should be `'yx'` here as `get_frame` returns a frame with only y, x axes.

pims/base_frames.py
```diff
@@ -606,3 +606,28 @@ def __repr__(self):
             s += "Axis '{0}' size: {1}\n".format(dim, self._sizes[dim])
         s += """Pixel Datatype: {dtype}""".format(dtype=self.pixel_type)
         return s
+
+    def _init_axis_if_exists(self, axis, size, min_size=1):
```
caspervdw: I think this function does not belong in the baseclass, the logic of how to init axes is really reader-specific.

Utils functions tend to grow out of hand in my experience. Just hardcode the logic if its simple.

pims/base_frames.py
```diff
@@ -606,3 +606,28 @@ def __repr__(self):
...
+        if size >= min_size:
+            self._init_axis(axis, size)
+
+    def _guess_default_iter_axis(self):
+        """
+        Guesses the default axis to iterate over based on axis sizes.
+        Returns:
+            the axis to iterate over
+        """
+        priority = ['t', 'z', 'c', 'v']
```
caspervdw: This type of 'guessing what the user wants' does not have my preference. I'd like this logic to be in the subclass

pims/imageio_reader.py
```diff
@@ -39,6 +46,23 @@ def __init__(self, filename, **kwargs):
...
+        self._init_axis_if_exists('t', self._len)
+
+        if len(self.sizes) == 0:
+            raise EmptyFileError("No axes were found for this file.")
+
+        # provide the default
+        self.iter_axes = self._guess_default_iter_axis()
+
     def get_frame(self, i):
         frame = self.reader.get_data(i)
```
caspervdw: get_frame should accept a dictionary of coords (like `t=2`)

caspervdw:
Thanks @rbnvrw  for this work, I left some comments but generally I really like the movement towards `FramesSequenceND`

rbnvrw:
Sorry @caspervdw I lost track of this PR.
I've implemented your changes.
Two of the builds (Python 2.7) on Travis pass, but for the other two (Python 3.6) I get this error:
```
RemoveError: 'setuptools' is a dependency of conda and cannot be removed from

conda's operating environment.

The command "conda update --yes conda" failed and exited with 1 during .
```
I guess this has something to do with the build environment config, but I couldn't find the cause.
Do you have any ideas?

rbnvrw:
Hi @caspervdw if you have the time, could you please look at this again? The tests are passing now.

caspervdw:
It looks good! Thanks for following up on this

One more general question, do we have to warn users for API changes of the ImageIOReader?

rbnvrw:
Hi @caspervdw thanks for looking at this, that is a good point!
What is the 'general practice' for pims? I don't see any changelog or something similar.

Should we put it in the description of the new release on GitHub or is that not visible/discoverable enough?
Or did you have something else in mind?

nkeim:
@rbnvrw As you suggested, please propose a note about this by adding it to #345. Or just post the text here and I can do it. Thanks!

rbnvrw:
@nkeim thanks, I have added a note to #345. Please check if it is OK and merge this one if you agree.

nkeim:
That’s good! Are there any new capabilities or applications this PR makes possible, or is it laying a foundation for the future?

rbnvrw:
The main thing is that it allows you to do axis bundling, iteration over (multiple) custom axes and setting default coordinates, as described here: https://soft-matter.github.io/pims/v0.4/multidimensional.html

nkeim:
Thanks! I think your release note undersells this enhancement :). I just pushed an edit to #345.

PR #321, opened 2019-04-06
Travisfixes

caspervdw:
This fixes two issues related to skimage and imagio changes:

 - a certain combination of skimage and numpy raises an `ImportError` on `import skimage.io` and not on `import skimage`. This makes the logic in that area a bit more robust
 - Newer imageIO versions return Inf on `reader.get_length`, you need to call `reader.count_frames` to get the length. This solves the `TypeErrors` encountered in https://github.com/soft-matter/pims/pull/317

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
     - python: "2.7"
       env: DEPS="numpy=1.8* slicerator tifffile" BUILD_DOCS=false
     - python: "2.7"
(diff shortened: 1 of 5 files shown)


caspervdw:
Also, as of version 2.5 imageIO does not support the ffmpeg on py2.7. So I pinned imageio to 2.4 in the tests.

nkeim:
Thanks! I was actually starting to work on a fix for this :). Your approach is very user-friendly, which is great.

PR #322, opened 2019-05-07
Fix PyAVReaderIndexed for videos with audio.

keunhong:
This PR introduces four changes:

1. The built-in PyAV .to_ndarray() function is used instead of a custom one which fixes some data format issues. I also imagine it reduces some overhead since it's a Cython function.
2. Skip non-video packets when indexing/seeking the container. This fixes an issue where reading some video files with an audio stream would cause errors.
3. Add a 'toc' argument to PyAVReaderIndexed so that pre-comptued tables can be passed in.
4. Removed an assertion that j == frame.index. As discussed in https://github.com/mikeboers/PyAV/issues/33, it seems like the index property may not be reliable in some cases. I've encountered files with packets with duplicate "index" values but the correct number of frames overall.

### DIFF
--- a/pims/pyav_reader.py
+++ b/pims/pyav_reader.py
@@ -308,10 +308,10 @@ class PyAVReaderIndexed(FramesSequence):
         return {'mov', 'avi',
                 'mp4'} | super(PyAVReaderIndexed, cls).class_exts()
 
-    def __init__(self, file, toc=None):
+    def __init__(self, file, toc=None, format=None):
         self.file = file
 
-        container = av.open(self.file)
+        container = av.open(self.file, format=format)
 
         # Build a toc
         if toc is None:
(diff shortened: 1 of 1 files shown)


keunhong:
Hi, is anyone willing to take a look at this PR?

pims/pyav_reader.py
```diff
@@ -22,12 +23,12 @@ def available():
     return av is not None
 
 
-def _to_nd_array(frame):
-    if frame.format.name != 'rgb24':
-        frame = frame.reformat(format="rgb24")
-    frame_arr = np.frombuffer(frame.planes[0], np.uint8)
-    frame_arr.shape = (frame.height, frame.width, -1)
-    return frame_arr
+def _next_video_packet(container_iter):
```
nkeim: This appears to fix a bug I also encountered (and tried to fix in a much less elegant way).

PR #301, opened 2018-06-29
Bump loci_tools to 6.4.0

caspervdw:

PR #324, opened 2019-05-14
Fix Bio-Formats download zips

joshmoore:
The zips used for testing Bio-Formats have been moved
to the scif.io domain. With this change, both 5.7.0
as well as 5.8.2 from gh-301 pass for:

```
nosetests --nologcapture -i bioformats
```

### DIFF
--- a/download_bioformats_test.py
+++ b/download_bioformats_test.py
@@ -15,19 +15,23 @@ def download_bioformats_tstfiles():
 
     def get_bioformats_file(filename, filepath='', url=''):
         if url == '':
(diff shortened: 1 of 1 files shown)


caspervdw:
Thanks! This script does not run on Travis, we used it only in the past for offline testing.

joshmoore:
Understood. The new URLs should still be good to go though.

joshmoore:
Is there anything else I should do here?

caspervdw:
Thanks for the reminder, merging this

PR #325, opened 2019-05-14
Dev Dockerfile

joshmoore:
Conda environment in Docker that matches the set up
in .travis.yml. At the moment, an interactive bash
shell is needed for jpype to properly find libjvm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+## Dockerfile for pims
+## -------------------
+##
+## By default, starts a bash shell:
+##
+##    docker build -t pims .
+##    docker run -ti --rm pims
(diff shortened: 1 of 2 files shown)


jthetzel:
Interesting
```
docker run -it --rm pims /bin/bash
nosetests
```
Succeeds.

But 
```
docker run -it --rm pims /bin/sh
nosetests
```
Fails.

$JVM_HOME is unset in `/bin/sh`, but correctly set in `/bin/bash`.

joshmoore:
@jthetzel, good point. The configuration of the conda environment would need to be updated to support non-bash shells:

```
RUN echo "source activate pims" >> ~/.bashrc
```

Issue #331, opened 2019-10-16
Docs: intersphinx


moi90:
It would be great if PIMS supported intersphinx:

http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

Then people could more easily link the documentation.

I did not find a http://soft-matter.github.io/pims/stable/objects.inv, so I assume that intersphinx has to be enabled to generate this file.

rbnvrw:
Thank you for the suggestion, that looks certainly interesting.
I'd be open to a PR if you have the time to integrate this into PIMS.

@nkeim What do you think of this suggestion? Is it possibly also interesting to add to Trackpy?

moi90:
It seems that objects.inv is build unconditionally, independently of an activated intersphinx extension. Could you search for *.inv files in your documentation build? Which version of sphinx do you use?

moi90:
Ok, I managed to find the latest objects.inv:

https://github.com/soft-matter/soft-matter.github.io/blob/master/pims/v0.4.1/objects.inv

My assumption was, that `/pims/stable` would be a link to the latest directory. But instead, the `/stable/index.html` contains redirect. I can see no code for that in `.travis.yml`. Do you do that manually? Could you use a symbolic link instead, so that `/stable/objects.inv` would work?

moi90:
Next problem: I'd like to link PyAVReaderIndexed. But because you do not include class documentation, I can't. I will create a pull request for including class documentation.

PR #334, opened 2019-10-17
DOC: Improve documentation of video.

moi90:
The generated documentation now includes the class documentation of PyAVReaderTimed,  PyAVReaderIndexed, and ImageIOReader. This helps with #331 as these classes are now included in objects.inv.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 build/*
 dist/*
 pims.egg-info/*
+PIMS.egg-info/*
 pims/version.py
 pims/loci_tools.jar
 examples/.ipynb_checkpoints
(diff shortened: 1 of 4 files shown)


moi90:
Thanks for merging!

My question is: Because of how `api.not_available` is used in `api`, it seems that you intended users to use the toplevel name of a class (`pims.PyAVReaderTimed` vs `pims.pyav_reader.PyAVReaderTimed`). Is that right?

Otherwise, I would reference the submodule name (`pims.pyav_reader.PyAVReaderTimed`) in the documentation. This would also allow to render documentation for classes need modules that are not installed in the environment for building the documentation.

What do you think?

PR #336, opened 2020-02-10
Add license to setup.py

sergiuser1:


rbnvrw:
Thank you for the suggestion, it is a good idea so I'll merge it.

Issue #318, opened 2019-01-17
Update Bioformats


joaomamede:
We have a GE/Deltavision Ultra microscope in the lab and while the newest command line Bioformats works, pims still uses an old version.
Is there anyway of me updating the bioformats loci jar to the newest bioformats version?

Thanks,
JM

GenevieveBuckley:
## Option 1 (quick to try)
In python, using the version of pims you have already:

```python
import pims
pims.bioformats._download_jar(version='6.0.1')  # the version number should be in quotes like this
pims.Bioformats('your_filename')
```

From the [list of bioformats versions](https://downloads.openmicroscopy.org/bio-formats/) it seems version 6.0.1 is the latest.


## Option 2 (also quick)
Instead of installing pims with the usual:
`pip install pims`

You could try:
`pip install git+https://github.com/caspervdw/pims.git@bump-bioformats`

This would give you loci_tools.jar version 5.8.2 (@caspervdw has an open pull request for this, so eventually this will make it into the master branch https://github.com/soft-matter/pims/pull/301)

## Option 3 (more complicated)
1. Make a fork of the pims repository (click the 'Fork' button in the top right hand corner of the pims page on github)
2. Make a new branch for your changes (you can do this by clicking the button 'branch: master' that is on the left hand side above the list of files in your repo. You could call the new branch `locitools`)
3. Edit line 57 of the file `pims/pims/bioformats.py`. Change the version number in quotes to the one you want:
```python
def _download_jar(version='5.7.0'):
```
4. Then reinstall pims with your new changes:
`pip install git+https://github.com/joaomamede/pims.git@locitools`

More generically, that's:
`pip install git+https://github.com/$YOUR_GITHUB_NAME/$GITHUB_REPO_NAME.git@$BRANCH_NAME`

joaomamede:
Hi, Thank you for the help.

I tried the first option (that would work for me), however.
 `In [1]: import pims

In [2]: print pims.__version__
0.4.1

In [3]: pims.bioformats._download_jar(version='6.1.1')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-8b7609da9818> in <module>()
----> 1 pims.bioformats._download_jar(version='6.1.1')

AttributeError: 'module' object has no attribute '_download_jar'


In [4]: 
`

I guess I'll create a fork in github to try.

joaomamede:
[...]

PR #346, opened 2020-05-01
Update bioformats docs

GenevieveBuckley:
After PR https://github.com/soft-matter/pims/pull/301 is merged & the default bioformats version bumped to 6.4, it would be good to have the newly exposed functionality documented.

This PR adds a short section to `docs/source/bioformats.rst` explaining how a user can update the version of bioformats/loci_tools.jar they are using with pims.

Relevant issue: https://github.com/soft-matter/pims/issues/318

### DIFF
--- a/doc/source/bioformats.rst
+++ b/doc/source/bioformats.rst
@@ -94,3 +94,18 @@ To update the version of bioformats you are using in pims:
     pims.bioformats.download_jar(version='6.4')
 
 Now you should be able to use pims with the updated bioformats version.
+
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
FYI: I was in two minds about including the extra note at the bottom on troubleshooting. I'm erring on the side of it being more helpful to have in there, given the eventual resolution to the discussion in https://github.com/soft-matter/pims/issues/318

rbnvrw:
This looks good to me. Lets wait for PR #301 to be updated and merged.

PR #349, opened 2020-05-06
Fix broken links in bioformats.rst

GenevieveBuckley:
A couple of the links in the pims bioformats documentation are broken, this PR should fix that.

Also, to try and avoid confusion I've made the bioformats version number in the download_jar() example match the current pims default.


rbnvrw:
Nice catch @GenevieveBuckley and thanks again for your contributions to pims!

PR #352, opened 2020-06-02
DEP: Pin bioformats at 6.x

nkeim:
This implements @joshmoore 's suggestion in #301 to pin to the bioformats major version, 6.x .


rbnvrw:
Nice, thanks! :+1:

PR #357, opened 2020-06-08
Add complete list of supported bioformats filetypes

GenevieveBuckley:
I was trying to open a deltavision `.dv` image file the other day and found that the bioformats reader doesn't recognise all bioformats filetypes. This means that the `pims.open()` command won't call out to the bioformats reader.

Even ahough you can force it to open correctly by calling `pims.Bioformats()` instead of `pims.open()`, this isn't ideal for downstream libraries/users (eg: dask-image uses `pims.opne()` and we wouldn't want to implement any extra logic around trying to pick the pims reader ourselves, since pims does that much better).

I got the list of bioformats supported file extensions from https://docs.openmicroscopy.org/bio-formats/5.7.1/supported-formats.html

There is some overlap with file extensions recognized by other readers, but that was true of the last set too & it shouldn't matter since I think the Bioformats reader is tried last.


nkeim:
This is a good idea! I would love to see it in v0.6.

GenevieveBuckley:
Note: this PR needs to specify a *relative* path to `bioformats_filetypes.txt`, then the CI should pass

tacaswell:
"power-cycled" to get new CI.

@GenevieveBuckley Is this list still accurate (2 years later)?

nkeim:
Looks like this code needs to find `bioformats_filetypes.txt` via the module path. Once that's done I'm for merging; even if the list is a bit outdated it's still a big enhancement.

GenevieveBuckley:
> "power-cycled" to get new CI.
> 
> @GenevieveBuckley Is this list still accurate (2 years later)?

I don't know, I haven't looked at it since then. 

You could cross check it against the latest (or whichever version pims currently pins to, which is not the one I originally looked at anymore I don't think) https://docs.openmicroscopy.org/bio-formats/5.7.1/supported-formats.html

Alternatively, perhaps it would be better to link directly to that page, either in addition to or instead of maintaining a separate copy of the list here. That would reduce the maintenance burden, although I guess dead links are always a possibility.

tacaswell:
Given that this is tied to a version of bioformats and #403 is changing that, I think we should do both together. 

I do not think we want have a network hit on import, it is probably best to just embed this list as a (very long) literal in the source (the test failures are failing to open the txt file, not sure if it is cwd related or packaging related...) .

PR #358, opened 2020-06-08
Allow imageio-ffmpeg to be optional

GenevieveBuckley:
I'm proposing making `imageio-ffmpeg` optional - if you don't have it installed you can still use regular `imageio` formats, and if you do have it then you get access to the extra movie file formats that are supported.

@jakirkham and I have been discussing whether to include required dependencies in our dask-image project so that image reading with pims works out of the box, even if installed into a new environment. Currently there isn't a lightweight set of dependencies for pims that covers basic image reading. Both `scikit-image` and `matplotlib` are pretty heavyweight so we don't want to require these, but we do want something relatively bare-bones.

We think that `tifffile` and `imageio` combined cover most of what our users want. But it turns out that the pims `imageio_reader.py` requires you to also install `imageio-ffmpeg` in order to use any of the imageio reader functionality - and that's also a heavyweight. If it was optional, that would solve things pretty neatly for us.


nkeim:
This seems very reasonable! Would it make sense to add the following environment to `.travis.yml`:

```yaml
    - python: "3.6"
      env: DEPS="numpy slicerator tifffile imageio" BUILD_DOCS=false
```

jakirkham:
I think we need to update [this check]( https://github.com/soft-matter/pims/blob/731203c3a2a2db20b93825cadef70a2c4dfcd6cc/pims/tests/test_common.py#L452 ). Also it looks like [one test needs `Jinja2`]( https://travis-ci.org/github/soft-matter/pims/jobs/699557715#L1141 ).

GenevieveBuckley:
Thanks John! I've updated the tests so we skip anything that needs to be skipped because of unavailable imports.

Note we still can't open a list of png files using only imageio (the pims ImageSequence reader requires either scikit-image or matplotlib), but this was true before as well. I'll open a separate issue for that since I think it would be a useful feature.

jakirkham:
Thanks Genevieve! 😄 LGTM

@nkeim, could you please take another look? 🙂

rbnvrw:
Thank you very much! Looks good. I'd merge it but maybe we want to wait for @nkeim 's input first.

jakirkham:
Thanks all! 😄

Issue #355, opened 2020-06-07
Update pims documentation & code re: deprecated scipy.ndimage.imread


GenevieveBuckley:
The [pims installation docs](https://github.com/soft-matter/pims/blob/master/doc/source/install.rst) say that you need either scikit-image, matplotlib, or scipy if you want to do basic image reading. This isn't true anymore, scipy [deprecated `imread` in 2017](https://github.com/scipy/scipy/pull/7869) and should be removed from that list.

Pims also still refers to the `scipy.ndimage.imread` function [in line 33 of image_sequence.py](https://github.com/soft-matter/pims/blob/cd40d4edcf1379d800e03273f8ffd63e097ec0c6/pims/image_sequence.py#L33).

nkeim:
Thank you @GenevieveBuckley ! Even though I was hoping that we would not make any more changes before v0.5, it is the perfect opportunity to fix this, because removing scipy support will no doubt break some existing code.

PR #345, opened 2020-05-01
DOC: Update release notes for v0.5

nkeim:
This lists notable changes as of today (April 30). Please point out omissions.

PR #359, opened 2020-06-15
Remove scipy

nkeim:
This addresses #355 by removing the fallback of ImageSequence to `scipy.misc.imread()`, which was deprecated and is now removed. Since this was the only use of `scipy` in PIMS, this is (nearly) equivalent to removing all references to scipy in the package.


conda-recipe/meta.yaml
```diff
@@ -15,15 +15,15 @@ requirements:
     - setuptools
```
caspervdw, line 15: This whole file is not used anymore - it is maintained at conda-forge now

nkeim:
This should now be ready to merge! I added mention of this PR to the v0.5 release notes in #345 .

Issue #356, opened 2020-06-08
tifffile no longer included in scikit-image


jakirkham:
scikit-image stopped vendoring tifffile with PR ( https://github.com/scikit-image/scikit-image/pull/4235 ). As of scikit-image 0.17.0, it is no longer included. Would be good to drop [fallback paths like this]( https://github.com/soft-matter/pims/blob/83be3b55e4d8f5d2b6ee4f2bc0ee167d46e45c07/pims/tiff_stack.py#L26 ) since they will always fail even when scikit-image is installed going forward.

cc @GenevieveBuckley (for awareness)

PR #362, opened 2020-06-19
Remove fallback path to deprecated skimage.external.tifffile

GenevieveBuckley:
Closes https://github.com/soft-matter/pims/issues/356

As of October 2019 `scikit-image` no longer vendors `tifffile`, so this fallback path to the deprecated `skimage.external.tifffile` will always fail from `skikit-image` version 0.17 onwards.

### DIFF
--- a/pims/tiff_stack.py
+++ b/pims/tiff_stack.py
@@ -22,10 +22,7 @@ except ImportError:
 try:
     import tifffile
 except ImportError:
-    try:
-        from skimage.external import tifffile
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks @GenevieveBuckley! Looks good 😄

@nkeim, do you have any thoughts on this? 🙂

nkeim:
Looks good! The instructions already tell everyone to install tifffile for TIFFs. Thanks for catching this.

Issue #363, opened 2020-06-19
eImageSequence reader 


GenevieveBuckley:

GenevieveBuckley:
Note: I did some initial experimenting here, and when I added a fallback path to the imageio imread that broke a lot of the tests - will need to look into this some other time.

```python
# pims/pims/image_sequence.py

# ...
has_skimage = False
try:
    from skimage.io import imread
    has_skimage = True
except ImportError:
    try:
        from matplotlib.pyplot import imread
    except ImportError:
        try:
            from imageio import imread
        except ImportError:
            imread = None
# ...
```

rbnvrw:
This would be nice to have indeed. Maybe this helps? https://imageio.readthedocs.io/en/stable/scipy.html Without knowing which tests fail, I'm not sure if I can be of help, but let me know if you need any.

nkeim:
I tried replacing the import section that @GenevieveBuckley listed above with

```python
has_skimage = False
from imageio import imread
```

and the ImageSequence tests pass. So there must be something I'm missing. @GenevieveBuckley do you want to create a PR?

GenevieveBuckley:
Ok, you made me wonder if it was just some weirdness with my local environment so I've made a PR. A build log with the errors is here: https://travis-ci.org/github/soft-matter/pims/jobs/703002972

It's the environment from this section of the `.travis.yml` matrix:
```
    - python: "3.8"
      env: DEPS="numpy slicerator tifffile imageio" BUILD_DOCS=false
```

PR #365, opened 2020-06-29
Extending ImageSequence reader when only imageio (but not skimage/matplotlib) is installed

GenevieveBuckley:
Closes https://github.com/soft-matter/pims/issues/363

Adds a fallback path to imageio imread for `pims/image_sequence.py`

This is a "nice to have" extension of the current capabilities, aimed at allowing a lightweight minimal-ish set of pims dependencies that can cover most of the basic use cases (eg `tifffile` and `imageio`) without requiring more heavyweight dependencies like `scikit-image`/`matplotlib` are also installed.

### DIFF
--- a/pims/image_reader.py
+++ b/pims/image_reader.py
@@ -25,7 +25,10 @@ except ImportError:
                  "in your script.)")
         warnings.warn(RuntimeWarning(ski_preferred))
     except ImportError:
-        imread = None
+        try:
+            from imageio import imread
+        except ImportError:
+            imread = None
 
 
 class ImageReader(FramesSequence):
--- a/pims/image_reader.py
+++ b/pims/image_reader.py
@@ -12,21 +12,22 @@ try:
     from skimage.io import imread
 except ImportError:
     try:
-        from matplotlib.pyplot import imread
-        # imread() works differently between scikit-image and matplotlib.
-        # We don't require users to have scikit-image, 
-        # but if we fall back to matplotlib, make sure the user 
-        # is aware of the consequences.
-        ski_preferred = ("PIMS image_reader.py could not find scikit-image. "
-                 "Falling back to matplotlib's imread(), which uses floats "
-                 "instead of integers. This may break your scripts. \n"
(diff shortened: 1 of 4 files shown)


nkeim:
The problem is in `image_reader.py`, which defines a PIMS class to read a single frame (I forget why we have this). Once both `image_sequence.py` and `image_reader.py` are updated we should be good to go!

GenevieveBuckley:
> Once both `image_sequence.py` and `image_reader.py` are updated we should be good to go!

This is the build log after adding an imageio fallback path to `image_reader.py`: https://travis-ci.org/github/soft-matter/pims/jobs/703756047

I haven't dug deeply into the errors yet, but I think it's something to do with the `plugin=` kwarg being passed into imread.

GenevieveBuckley:
Also, should we consider making the order of preference for `imread` go: scikit-image > imageio > matplotlib, given the warning about "imread() works differently between scikit-image and matplotlib? ["We don't require users to have scikit-image, but if we fall back to matplotlib, make sure the user is aware of the consequences."](https://github.com/soft-matter/pims/blob/3c8f7832113517a37ddd65a2803a6b17f9c33e4c/pims/image_reader.py#L16) Any opinion here @nkeim ?

GenevieveBuckley:
FYI: `TestImageSequenceNaturalSorting` is failing locally if `imageio`  imread is used (because of the `plugin=` kwarg).

I've pushed an update to skip this test class for now while we sort the other stuff out, but this either needs to be fixed, or for a new test to be added covering the same functionality with imageio.

nkeim:
Great catch, @GenevieveBuckley ! Yes, assuming that imageio's `imread` is more similar to scikit-image's, we definitely want to prefer it over matplotlib's.

tacaswell:
I think this is obsoleted by #392 which removes Matplotlib as a reader all together and adds a hard dependency on imageio.


Thank you for this @GenevieveBuckley and sorry we did not get it merged 2 years ago.

Issue #374, opened 2020-10-22
Value error opening tiff


jrussell25:
Hi there, ive been getting a value error having to do with datetime when pims looks for metadata in some microscopy images I  have. This could be related to #294 but wasn't obviously the same. Below is the code and [here is a link to the test image](https://github.com/jrussell25/data-sharing/blob/master/Pos001_S001_t21_z4_ch01.tif)

```
import pims
import tifffile as tiff
print(pims.__version__) #0.5
print(tiff.__version__) #2020.10.1

tiff_test = tiff.imread('Pos001_S001_t21_z4_ch01.tif') #Its a phase contrast image of yeast cells

pims_test = pims.open('Pos001_S001_t21_z4_ch01.tif')

im = pims_test.get_frame(0)
```

The last line raises 

`ValueError: invalid literal for int() with base 10: '0:00'`

<details><summary>Command Line Output</summary>
<pre>
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-173-3ac3cb358dd8> in <module>
----> 1 im.get_frame(0)

/opt/anaconda3/envs/mpl/lib/python3.8/site-packages/pims/tiff_stack.py in get_frame(self, j)
    119         t = self._tiff[j]
    120         data = t.asarray()
--> 121         return Frame(data, frame_no=j, metadata=self._read_metadata(t))
    122 
    123     def _read_metadata(self, tiff):

/opt/anaconda3/envs/mpl/lib/python3.8/site-packages/pims/tiff_stack.py in _read_metadata(self, tiff)
    150                 md[key] = ''
    151             if key == 'DateTime':
--> 152                 md[key] = _tiff_datetime(md[key])
    153         return md
    154 

/opt/anaconda3/envs/mpl/lib/python3.8/site-packages/pims/tiff_stack.py in _tiff_datetime(dt_str)
     49 def _tiff_datetime(dt_str):
     50     """Convert the DateTime string of TIFF files to a datetime object"""
---> 51     return datetime(year=int(dt_str[0:4]), month=int(dt_str[5:7]),
     52                     day=int(dt_str[8:10]), hour=int(dt_str[11:13]),
     53                     minute=int(dt_str[14:16]), second=int(dt_str[17:19]))

ValueError: invalid literal for int() with base 10: '0:00'
</pre>
</details>

The `pims.open` seem fine and it gets the frame size and pixel type correctly, the problem only occurs when trying to get the data into an array.

Thanks for the help.

[...]

PR #375, opened 2020-10-26
Fix for issue 374, image with invalid datetime tag

GenevieveBuckley:
Closes https://github.com/soft-matter/pims/issues/374

Using [this test image](https://github.com/jrussell25/data-sharing/blob/master/Pos001_S001_t21_z4_ch01.tif), you can now do this:
```python
import pims
img = pims.open('Pos001_S001_t21_z4_ch01.tif')
img.get_frame(0)  # no error
```

### DIFF
--- a/pims/tiff_stack.py
+++ b/pims/tiff_stack.py
@@ -1,7 +1,10 @@
 import os
 from datetime import datetime
 import itertools
+import warnings
+
 import numpy as np
+
 from pims.frame import Frame
 
 try:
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Some of the CI tests are failing, I believe these are unrelated to this PR

pims/tiff_stack.py
```diff
@@ -146,7 +146,10 @@ def _read_metadata(self, tiff):
             except UnicodeDecodeError:
                 md[key] = ''
             if key == 'DateTime':
-                md[key] = _tiff_datetime(md[key])
+                try:
+                    md[key] = _tiff_datetime(md[key])
+                except Exception:
```
tacaswell, line 151: Can we be a more selective with the exceptions we suppress here?

It would be good at also issue a `warning.warning` (as I don't think we have threading proper logging through pims yet).

GenevieveBuckley: Good suggestions 👍

PR #383, opened 2021-02-13
Drop dependency on nose for tests; fix numpy and python deprecations.

anntzer:
See https://github.com/soft-matter/trackpy/pull/588 for rationale of dropping nose.


nkeim:
Splendid! Thanks @anntzer !

PR #391, opened 2021-07-31
Remove dependency on pypandoc in setup.py.

anntzer:
markdown is natively supported since setuptools 36.4.0 (2017), and
anyways that's only a version requirement for the person doing the
release, not for anyone else (older setuptools just ignore the kwarg
with a warning).

Closes https://github.com/soft-matter/pims/pull/361.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,13 +9,6 @@ try:
 except IOError:
     descr = ''
 
-try:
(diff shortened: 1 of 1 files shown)


nkeim:
It's great to get more cobwebs out. Thanks @anntzer !

PR #392, opened 2021-07-31
Add dependency on imageio; remove matplotlib-based reader.

anntzer:
matplotlib's imread is now anyways a thin shim around pillow (a
dependency of matplotlib), so if you have matplotlib installed, you may
as well use imageio's imread: imageio is a smaller dependency, and has
more consistent behavior (it doesn't return pngs as floats).

imageio is also a dependency of scikit-image (since 0.15), so it seems
reasonable to directly depend on it instead.  For now, the scikit-image
reader was kept, as removing it (possibly in the future) would entail
removing support for the `plugin` kwarg, and thus an API break.
(See also https://github.com/soft-matter/pims/pull/273#discussion_r141855815.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ Formats readable by pims include:
 
 PIMS is based on readers by:
 * [scikit-image](http://scikit-image.org/)
-* [matplotlib](http://matplotlib.org/)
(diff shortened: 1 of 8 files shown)


tacaswell:
@anntzer needs a rebase.

anntzer:
Done.

PR #395, opened 2021-08-28
Update pyav_reader.py to make generator threadsafe

BakudanKame:
I had an issue where code using pyav_reader.py would randomly break and throw a "ValueError: Generator Already Executing Error"; the stack indicated that the function _gen_frames was the culprit; I added a function and a decorator (line for line copied from the sources below with one change, re-naming "next" to "__next__") that makes a non-threadsafe generator threadsafe. This resolved the issue on both Windows and Mac. I thought I'd send it over to you guys for your perusal! Thanks for the great package!

The original error message:  Screen Shot 2021-08-28 at 3 40 56 PM

Sources:
https://anandology.com/blog/using-iterators-and-generators/
https://stackoverflow.com/questions/41194726/python-generator-thread-safety-using-keras

### DIFF
--- a/pims/pyav_reader.py
+++ b/pims/pyav_reader.py
@@ -27,6 +27,28 @@ def _next_video_packet(container_iter):
     raise ValueError("Could not find any video packets.")
 
 
+class threadsafe_iter:
+    """Takes an iterator/generator and makes it thread-safe by
+    serializing call to the `next` method of given iterator/generator.
+    """
+    def __init__(self, it):
+        self.it = it
+        self.lock = threading.Lock()
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        with self.lock:
+            return self.it.next()
+          
+def threadsafe_generator(f):
+    """A decorator that takes a generator function and makes it thread-safe.
+    """
+    def g(*a, **kw):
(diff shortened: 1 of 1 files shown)


pims/pyav_reader.py
```diff
@@ -2,6 +2,7 @@
                         unicode_literals)
 
 import numpy as np
+import threading
```
BakudanKame, line 5: Forgot to add the import for threading to the original pull request!

tacaswell:
"powercycled" to get CI.

pims/pyav_reader.py
```diff
@@ -27,6 +28,28 @@ def _next_video_packet(container_iter):
...
+        return self
+
+    def __next__(self):
+        with self.lock:
+            return next(self.it)
+          
+def threadsafe_generator(f):
+    """A decorator that takes a generator function and makes it thread-safe.
+    """
+    def g(*a, **kw):
```
tacaswell, line 49: Can you add `functools.wraps(f)` here as well?

tacaswell:
@BakudanKame Sorry for the very slow review!

This seems reasonable, but I am a bit wary of tacking locks onto things, is there a reasonable way to let the user opt-into the locking/thread safety?

Issue #393, opened 2021-08-03
Migration to GitHub actions needed


nkeim:
Travis CI no longer has a sustainable free plan, so pims pull requests are no longer being tested automatically. A migration to GitHub actions is needed, which will likely be easier than the heroic work by @caspervdw in https://github.com/soft-matter/trackpy/pull/640 .

The latest PR's by @anntzer look great (thanks!), but I'm in favor of waiting until we get CI back, especially for #392 which should go with a reference build environment.

anntzer:
Closed by #437, afaict.

PR #399, opened 2021-10-29
modernize test suite + py311 support

tacaswell:
SafeConfigParser has been deprecated since Python 3.2 and will be removed in py311.

https://github.com/python/cpython/pull/28292
https://bugs.python.org/issue45173

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,24 @@
+name: Docs
+
+on: [push, pull_request]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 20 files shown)


nkeim:
Thanks @tacaswell ! This looks OK to merge. Ideally we can wait until we get CI tests working again… see #393 . But if 3.11 comes out first then we should just go for it.

tacaswell:
ok, I think I have this working, but need to do a slicerator release first.

tacaswell:
I'm going to merge this to get GHA going.  I expect there will be some follow up PRs to fix more things...

nkeim:
Thanks @tacaswell !!! 🎉 It's so good to have this repo back on a solid footing.

PR #407, opened 2022-04-07
Bld docs and config

tacaswell:


tacaswell:
Self-merging to un-break CI

PR #408, opened 2022-04-08
Drop six, drop some Py2 backcompat code.

anntzer:
The numpy>=1.19 bound effectively means Python2 is no longer supported.


tacaswell:
rebased due to conflicts in with #392 .

tacaswell:
It looks like imageio 2.16.2 (released yestreday) added warnings for migrating to a their v3 API (which looks nice! https://imageio.readthedocs.io/en/stable/reference/userapi.html#migrating-to-the-v3-api) but this is causing warnings in both our code and skimage.

anntzer:
Looks like we both rebased at the same time :)
The fix wrt imageio is probably just a bunch of version-dependent imports, but will depend on how far back you want to support (imageio.v2 only exists since imageio 2.16.0), so I'll you decide on that.

tacaswell:
"fixed" the failures by pinning imageIO back and created #411 to track the need to fix this.

PR #423, opened 2022-07-22
Clarify sorting rule in ImageSequence docs.

anntzer:
Just writing down what `pims.utils.sort.natural_keys` does.


doc/source/image_sequence.rst
```diff
@@ -14,6 +14,11 @@ filenames.
 * the filepath of a zipped archive, such as ``'my_directory/all-images.zip'``
 * a list of filepaths, such as ``['image1.png', 'image2.png']``
 
+In all cases except the last one, to sort the filenames, sorting keys are
```
nkeim, line 17: This is clear but has a typo ("are generated are sorted"). Would it also help to give a short example? You could say something like

```
['abc3.png', 'abc12.png', 'def3.png'] are in sorted order
```

anntzer: Sure, fixed & included example.

nkeim:
Merged. Thanks @anntzer for improving pims!!

PR #424, opened 2022-08-01
Be more lenient when parsing the TIFF DateTime metadata tag.

anntzer:
Although the TIFF spec does specify a fixed format for the DateTime
metadata tag, other formats are also found "in the wild", e.g. the
version of MetaMorph I have appears to use "20220101 00:00:00.000".
Obviously this isn't really something end users can do anything about.

The BioFormats library appears to allow for a variety of formats (that
cannot be confused for one another); reproduce their logic here.
(See DATE_FORMATS in BaseTiffReader.java and the handling of
milliseconds in DateTools.java's parseDate).

See:
https://github.com/ome/bioformats/blob/1cf1fefa364e19c8fb8e81d6cda0824f353dc4df/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java#L75-L82
https://github.com/ome/bioformats/blob/1cf1fefa364e19c8fb8e81d6cda0824f353dc4df/components/formats-bsd/src/loci/formats/in/BaseTiffReader.java#L431
https://github.com/ome/ome-common-java/blob/a77eadd32d66d0eacf7103ea70aa3def88e80a63/src/main/java/loci/common/DateTools.java#L238

### DIFF
--- a/pims/tiff_stack.py
+++ b/pims/tiff_stack.py
@@ -43,9 +43,34 @@ _dtype_map = {4: np.uint8,
 
 def _tiff_datetime(dt_str):
     """Convert the DateTime string of TIFF files to a datetime object"""
-    return datetime(year=int(dt_str[0:4]), month=int(dt_str[5:7]),
-                    day=int(dt_str[8:10]), hour=int(dt_str[11:13]),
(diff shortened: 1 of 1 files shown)


nkeim:
Thanks! Cross-referencing this with #374 and #375 .

PR #428, opened 2022-10-09
fix branch name in link to example nb

dmarx:


nkeim:
Thanks for the helpful and important fix!!

I'm going to merge this, but tests with the latest matplotlib are failing with `PendingDeprecationWarning('The set_tight_layout function will be deprecated in a future version. Use set_layout_engine instead.')`. I will create a separate issue about that.

PR #431, opened 2022-11-28
Fix seek in PyAVReaderTimed

VladyslavUsenko:
This PR fixes a seek method of the `PyAVReaderTimed`. 

The timestamp, that is computed in the `seek` method, corresponds to stream and not to the container. This PR updates `self._stream.container.seek` to seek in the stream. In my experiments, it significantly reduces the time for `get_frame` calls on large video files.

### DIFF
--- a/pims/pyav_reader.py
+++ b/pims/pyav_reader.py
@@ -226,7 +226,7 @@ class PyAVReaderTimed(FramesSequence):
         # the ffmpeg decode cache is flushed automatically
 
         timestamp = int(i / (self._frame_rate * self._stream.time_base))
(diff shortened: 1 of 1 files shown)


VladyslavUsenko:
Moreover, it looks like here seek is used correctly: https://github.com/soft-matter/pims/blob/main/pims/pyav_reader.py#L397

danielballan:
Thanks, @VladyslavUsenko.

I approved the CI to run, and I notice some failures that look unrelated to the change. I will take a closer look tomorrow.

VladyslavUsenko:
Hi @danielballan, any progress on this?

PR #434, opened 2023-01-24
Don't emit warnings that duplicate exceptions on opening failure.

anntzer:
The error messages will already be displayed in the exception, so no need to emit them as warnings too.  Moreover, warnings are somewhat more annoying to suppress than exceptions.

(One *could* consider still emitting the warnings if e.g. the first reader failed but the second one succeeds, but I'm not sure it's worth the trouble.)

(CI failure is due to skimage-latest-release not being compatible yet with numpy-latest-release; I can put up a PR pinning numpy if you'd like.)

### DIFF
--- a/pims/api.py
+++ b/pims/api.py
@@ -195,15 +195,14 @@ def open(sequence, **kwargs):
                 return 10
         return sorted(handlers, key=priority, reverse=True)
 
-    exceptions = ''
+    messages = []
     for handler in sort_on_priority(eligible_handlers):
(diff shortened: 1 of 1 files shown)


nkeim:
Thanks for this. I agree that the potential annoyance of these warnings could be high, and the value is relatively low in practice—especially since if you want to understand why a specific reader was passed over, you should just call it directly to find out what went wrong and get a full stack trace. Merging!

Issue #426, opened 2022-10-04
loci_tools.jar issue


perdigao1:
Example code

```
import pims
image = pims.bioformats.BioformatsReader(somefilename)
```

And error messages that spit out
```
UserWarning: loci_tools.jar not found, downloading
  warn('loci_tools.jar not found, downloading')

...

HTTPError: HTTP Error 404: Not Found

```

Note here:

File [...]\site-packages\pims\bioformats.py:78, in download_jar(version)
     74 url = ('http://downloads.openmicroscopy.org/bio-formats/' + version +
     75        '/artifacts/loci_tools.jar')

which points to https://downloads.openmicroscopy.org/bio-formats/6.10.1/artifacts/loci_tools.jar , but this file/url does not exist

Please fix

GenevieveBuckley:
You could try branch https://github.com/soft-matter/pims/pull/403 (using bioformats_package.jar instead of loci_tools.jar)

jstriebel:
@perdigao1 You can also download an older version of bioformats which still contains the loci_tools.jar. The newest working version seems to be `6.7.0`. We use such a workaround here:
https://github.com/scalableminds/webknossos-libs/blob/1d4831dd00862de7ee634137f5bbeaf0ba4c10e8/webknossos/webknossos/dataset/_utils/pims_images.py#L251-L258

Issue #436, opened 2023-01-24
Support pathlib.Path inputs


anntzer:
Currently pathlib.Path filenames are mostly not supported by pims (erroring in various places when passed as argument).  I can try to improve on that, but one question is whether they should be globbed or not (I would think *no*, but this is not entirely clear-cut, see e.g. https://github.com/python/cpython/issues/79634).

PR #437, opened 2023-01-24
Update/fix CI setup.

### LINKED ISSUE #411: Account for deprecation warnings in imageio
ImageIO 2.16.0 added a v2 and v3 name spaces and ImageIO 2.16.2 added warnings when using the top-level namespace which is causing our tests to fail.  As a stop-gap I have pinned the CI dependency back (#410 ), but we should properly fix this both here and in skimage.

I think 2.16.0 (which came out in Feb 22) is too new to depend on so we should do the "import the new one, fallback to the old one" dance.

### LINKED ISSUE #429: MNT: transition matplotlib code to set_layout_engine
Several tests with latest dependencies now fail with `PendingDeprecationWarning('The set_tight_layout function will be deprecated in a future version. Use set_layout_engine instead.')`

It looks like at this stage, we should avoid changing the code if possible since `set_layout_engine` was introduced only about a year ago. But now that the warning has done its job, we should change the tests to ignore it.

anntzer:
I went for the big omnibus PR...

- Move list of dependencies into GHA yaml files; restrict to strict minimum for CI, not including black, flake8, twine, etc.  (#393)
- Test both with no optional deps and with all optional deps, except pylibtiff (because it seems tricky to install on CI -- it tries to look for the tifffile header and fails to find it).  Note that the pylibtiff README explicitly states that tifffile is as efficient.
- Pass -ra to pytest to indicate why tests are skipped.
- Drop old, now-unused travis conf.
- Fix for API changes re: imageio.v3 (closes #411), matplotlib layout engines (closes #429, also works around https://github.com/matplotlib/matplotlib/pull/24971), pyav format long_name, jpype threads and buffers (now exposed using the buffer protocol, not numpy arrays).
- imageio now emits a warning if loading a tiff and tifffile is not separately installed (they deprecated their vendored copy of tifffile), let's just add tifffile as a dependency.
- Pin bioformats jar to 6.7.0 (last working version); pre-download it before running tests to avoid download warnings causing the test suite to fail.  (See #426; likely moving to bioformats_package.jar could still be a good idea?)
- AFAICT the bioformats reader needs to explicitly construct a memoryview around the data returned by openBytes, otherwise the underlying buffer can become dangling (this leads to brittle, sometimes failing tests on CI).
- Work around the upcoming deprecation of distutils by adding a dependency on packaging instead (technically it could be made optional as it is only needed for bioformats, but it is likely very widely installed, not too heavy, and always depending on it seems just simpler).
- Mark all tests that pass plugin=... as requiring skimage (otherwise they emit a warning that plugin=... is ignored).
- Explicitly close file at teardown of bioformats tests to avoid a ResourceWarning. Likewise, imageio-ffmpeg seems to have some reference cycle(?) which makes a gc.collect() sometimes necessary to avoid a ResourceWarning.

Unless there's strong reasons to do otherwise, I would also suggest dropping support for pylibtiff (and perhaps also just pick a side between pillow's tiff and tifffile); and likewise pick one movie reader (pyav seems favored by skimage https://scikit-image.org/docs/stable/user_guide/video.html#pyav, and has wheels so installation should be easy in practice).


nkeim:
Wow!!! What a gift for pims! Thanks @anntzer !!

I see no reason not to merge now, especially since this PR has had some time to season, and because we really could use working CI again. 

I see that you are in support of some further changes, which I summarize as
1. dropping `pylibtiff`
2. dropping `moviepy`
3. switching to bioformats_package.jar instead of loci_tools.jar

I see no problem with any of those given how much the package ecosystem has evolved since pims was first written, but I agree that those should be done in separate PRs since they would require some careful changes to the code, and of course there could be some users out there for whom these would be big problems. We already have #426 to discuss the bioformats change; should we open additional issues to discuss the first two?

anntzer:
I would also drop imageio-ffmpeg (per https://imageio.readthedocs.io/en/v2.25.1/_autosummary/imageio.plugins.ffmpeg.html#module-imageio.plugins.ffmpeg).

While I appreciate that this may be disruptive for some users, I simply notice that maintaining working CI for pims seems to have been challenging recently, and e.g. in the PR above I had to make various fixes for a number of readers, but I have no idea of whether the pylibtiff reader works as even installing that package properly is challenging (and I had limited interest in fixing that when the package README itself suggests that other tiff libraries may be better suited).  Also note that my interest in getting CI fixed is actually to help the review of my other PRs (right now only #434; I could perhaps also give a try at fixing #436 now that CI is fixed).

You can open separate issues to keep track of the decision process but I don't really have much to add on top of that.

PR #445, opened 2023-04-07
FIX: Close files in case of reader failure

### LINKED ISSUE #416: Failing tests in test_bioformats.py
Tests are failing in `test_bioformats.py`. I noticed this while working on PR https://github.com/soft-matter/pims/pull/403, but it seems to happen for all branches, so I don't think it's caused by my code.

To create a test environment, including dependencies for video files:
```
conda create -n pims-dev python=3.9 pip ipython
conda activate pims-dev
cd GitHub/pims
python -m pip install -r requirements-dev.txt
python -m pip install -e .
conda install -c conda-forge av imageio-ffmpeg moviepy tifffile
```

To run the tests:
```
python download_bioformats_test.py
pytest pims/tests/test_bioformats.py -vv
```

Problems encountered:
* Inconsistent failing tests. In `test_bioformats.py` I see a lot of `test_repr` failures, but when I run just the single test it passes fine. Commenting out `test_repr` leads to other tests failing (that again, pass fine when run individually). Is something weird happening where the test files are not being closed properly? I don't know how to debug this.
* There's something wrong with the norpix reader. It can't open the `HEART.SEQ` test image file. Oddly, all of the tests in `pytest pims/tests/test_norpix.py` pass. But `pytest pims/tests/test_bioformats::TestBioformatsSEQ::test_open` fails.

More weird stuff I have noticed:
* Only the PyAV reader can open the file `pims/tests/data/bioformats/wtembryo.mov`. If PyAV is not available, sometimes other handlers try and fail (which casues a test failure). Given that this file will play in VLC but not QuickTime, I assume that's something to do with the codecs? I don't think this is a problem, because at least *something* can open the file, but it does mean you must have PyAV installed as a dependency for the tests not to fail.

nkeim:
This fixes #416 by ensuring that files opened by the NorPix and cine readers are closed promptly if there's a problem during initialization (e.g. invalid file format).

### DIFF
--- a/pims/cine.py
+++ b/pims/cine.py
@@ -336,7 +336,6 @@ class Cine(FramesSequence):
     For a .chd file, this class only reads the header, not the images.
     
     """
-    # TODO: Unit tests using a small sample cine file.
     @classmethod
     def class_exts(cls):
         return {'cine'} | super(Cine, cls).class_exts()
@@ -742,6 +741,10 @@ class Cine(FramesSequence):
 
     def close(self):
         self.f.close()
+
+    def __del__(self):
+        if hasattr(self, 'f'):
+            self.f.close()
 
     def __unicode__(self):
         return self.filename
--- a/pims/imageio_reader.py
+++ b/pims/imageio_reader.py
@@ -149,3 +149,7 @@ class ImageIOReader(FramesSequenceND):
     def close(self):
         self.reader.close()
         super().close()
+
+    def __del__(self):
+        if hasattr(self, 'reader'):
+            self.reader.close()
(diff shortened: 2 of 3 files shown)


nkeim:
The problem in #416 was quite reproducible and really is fixed by this PR, but I've also seen similar sporadic failures with `TestVideo_ImageIO.test_simple_negative_index`. The second commit didn't seem to help.

GenevieveBuckley:
> The problem in #416 was quite reproducible and really is fixed by this PR

This is great!

> but I've also seen similar sporadic failures with `TestVideo_ImageIO.test_simple_negative_index`. The second commit didn't seem to help.

Perhaps it is worth opening an issue specifically for this? This PR already fixes one thing, I don't want to see it blocked because it doesn't fix *everything*.

PR #457, opened 2024-04-15
MNT: numpy2 compatibility

tacaswell:
Rather than chase private API imports, just copy it.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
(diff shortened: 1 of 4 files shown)


tacaswell:
testing with numpy2 won't work because we need numpy2 built versions of atleast skimage.  The tests pass with numpy2 for me locally.

tacaswell:
I slightly jumped to conclusions and only 3.9 failed with `--pre` https://github.com/tacaswell/pims/actions/runs/8694501189/job/23843551574 is a run where everything but 1 passes and the failure is skimage/numpy compilation issues (`ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject`)

danielballan (approved):

Issue #462, opened 2025-03-24
Error with latest setuptools version


DominicOram:
Trying to install `pims` with the latest `setuptools` produces the following error:
```
Obtaining file:///scratch/ffv81422/bluesky_work/pims
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/scratch/ffv81422/bluesky_work/pims/setup.py", line 41, in <module>
          setup(**setup_parameters)
        File "/scratch/ffv81422/bluesky_work/pims/.venv/lib/python3.11/site-packages/setuptools/__init__.py", line 116, in setup
          _install_setup_requires(attrs)
        File "/scratch/ffv81422/bluesky_work/pims/.venv/lib/python3.11/site-packages/setuptools/__init__.py", line 87, in _install_setup_requires
          dist.parse_config_files(ignore_option_errors=True)
        File "/scratch/ffv81422/bluesky_work/pims/.venv/lib/python3.11/site-packages/setuptools/dist.py", line 730, in parse_config_files
          self._parse_config_files(filenames=inifiles)
        File "/scratch/ffv81422/bluesky_work/pims/.venv/lib/python3.11/site-packages/setuptools/dist.py", line 599, in _parse_config_files
          opt = self._enforce_underscore(opt, section)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/scratch/ffv81422/bluesky_work/pims/.venv/lib/python3.11/site-packages/setuptools/dist.py", line 629, in _enforce_underscore
          raise InvalidConfigError(
      setuptools.errors.InvalidConfigError: Invalid dash-separated key 'description-file' in 'metadata' (setup.cfg), please use the underscore name 'description_file' instead.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
```

To reproduce:
* `pip install --upgrade setuptools`
* `pip install -e .`

jimmymathews:
I just want to mention a subtlety in case it is helpful for anyone encountering this, or helpful for the resolution of this issue.

[...]

PR #463, opened 2025-03-24
Fix `description_file` metadata

Zethson:
Fixes https://github.com/soft-matter/pims/issues/462

Requires an urgent release.


DominicOram:
You beat me to it by minutes! Thank you!

